PR #51, opened 2025-01-24
Fix type checking of Type::Bits const params in instantiations

keks:
When instantiating a package with Type::Bits in a game, we didn't compare the correct strings of the Bits.

Let's say we have a package with parameters `n: Int` and `k: Bits(n)`, and a game  with consts `m: Int` and `k: Bits(m)`, and int instantiates the package using `n: m` and `k: k`, then so far the type checker compared `Type::Bits("m") with `Type::Bits("n")`, which failed.

Conversely, if we had the same package and a game with constants `n: Int`, `m: Int`, `k: Bits(n)`, and we instantiated the package `n: m`, `k: k`, the typechecker would fail to identify that there was an issue.

So far we added tests for the first scenario, but not the latter. In fact, the latter turned up in the code a few times and we had to fix these tests as well.

### DIFF
--- a/example-projects/rosenpass-basic/games/KEMAssumption.game.ssp
+++ b/example-projects/rosenpass-basic/games/KEMAssumption.game.ssp
@@ -12,11 +12,11 @@ composition KemAssumption {
 
   instance kem = KEM {
     params {
-      kem_sk_len: kem_sk_len,
-      kem_pk_len: kem_pk_len,
-      kem_ct_len: kem_ct_len,
-      kem_shk_len: kem_shk_len,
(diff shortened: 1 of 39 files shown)


keks:
There still are missing error messages that currently just are `todo!()`s, and the rosenpass-basic example still don't work, which needs to be investigated.

keks:
Huh, the problem seems to be that we need to perform that mapping in other places too, which might not be as easy.

The alternative approach would be to just replace all instances of the old `Type::Bits` with one with the new string, but I am a bit concerned that this ends us up in a half-rewritten state.

PR #82, opened 2025-05-02
Feat/singlestepping

siccegge:


keks:
I think this looks pretty good.

I pushed two commits:
- one to format the source code
- one that slightly modifies the output:
  - equivalence becomes `==`, reduction becomes `~=`
  - these signs are aligned so they match that of the longest left name

I hope it was fine that I just pushed to your branch.

keks (approved):

siccegge:
Sure :upside_down_face:

PR #83, opened 2025-05-04
Feat/sample ids

### LINKED ISSUE #33: Unstable randomness counter in incremental pseudo-code
When slightly editing the pseudo-code, the randomness counters might change. This is a problem for incremental changes, since they require re-writing/checking of the invariant, i.e., revisiting all(!) randomness samplings which the invariant refers to.

siccegge:


keks (commented):
I think this looks good, but I would be a bit more confident if we had at least one test. The new syntax itself looks fine to me.

keks:
We discussed this today and the new idea is to replace the integer sample id in the SMT-LIB code with a type that holds three values:
- package instance name
- oracle name
- sample name
If the sample name is explicitly assigned, we use that. Otherwise, we use the string of the number in normal decimal numbers (usually "0" or "1" or so).

siccegge:
fallback ist immer noch der "globale" counter statt pro-orakel

siccegge:
we still use the old counter in some places ... e.g. as offsets into the sample positions returned as part of aux and also when accessing pkgstate. changing pkgstate would be relatively straight forward. Offsets we could replace by storing a `Vec<((PkgInstName,OracleName), Position>` instead of a `Vec<Position>` and searching.

@keks opinions?

src/writers/smt/patterns/datastructures/game_state.rs
```diff
@@ -39,7 +40,7 @@ impl<'a> NameSection for GameStateSelector<'a> {
             GameStateSelector::PackageInstance { pkg_inst_name, .. } => {
                 builder.push("pkgstate").push(pkg_inst_name)
             }
-            GameStateSelector::Randomness { sample_id } => builder.push("rand").push(sample_id),
+            GameStateSelector::Randomness { sample_pos } => builder.push("rand").push(sample_pos.sample_id),
```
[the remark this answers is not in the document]
siccegge: Should be addressed by e7349ab

src/writers/smt/exprs.rs
```diff
@@ -77,6 +78,24 @@ impl From<bool> for SmtExpr {
     }
 }
 
+impl From<&SamplePosition> for SmtExpr {
+    fn from(position: &SamplePosition) -> Self {
+        let sample_identifier = if let Some(sample_name) = &position.sample_name {
+            sample_name
+        } else {
+            &format!("{}", position.sample_id)
```
siccegge, line 86: this is unreachable

src/writers/smt/exprs.rs
```diff
@@ -77,6 +78,24 @@ impl From<bool> for SmtExpr {
...
+        };
+
+        (
+            "sample-id",
+            format!("\"{}\"", position.inst_name),
+            format!("\"{}\"", position.oracle_name),
+            format!("\"{}\"", sample_identifier),
+        )
```
siccegge, line 94: Into<SmtExpr> ?

src/parser/package.rs
```diff
@@ -934,9 +934,12 @@ pub fn handle_code(
                         name_ast,
                         oracle_name,
                         tipe.clone(),
-                    )
-                    ?;
-                    Statement::Sample(ident, None, None, tipe, full_span)
+                    )?;
+                    let sample_name = match inner.next() {
+                        None => None,
+                        Some(sample_name) => Some(sample_name.as_str().to_string()),
```
siccegge, line 940: this could be a map

src/transforms/samplify.rs
```diff
@@ -10,17 +10,19 @@ use std::iter::FromIterator;
...
     pub inst_name: String,
     pub pkg_name: String,
+    pub oracle_name: String,
 
     pub dst_name: String,
     pub dst_index: Option<Expression>,
 
     pub sample_id: usize,
     pub tipe: Type,
+    pub sample_name: Option<String>,
```
siccegge, line 25: This does not need to be an option!

src/writers/smt/contexts/game_inst.rs
```diff
@@ -201,12 +205,12 @@ impl GameInstanceContext<'_> {
 
     pub(crate) fn smt_eval_randfn<CTR: Into<SmtExpr>>(
         &self,
-        sample_id: usize,
+        sample_info: &SamplePosition,
```
keks, line 208: maybe call this sample_pos

src/writers/smt/writer.rs
```diff
@@ -638,9 +645,15 @@ impl<'a> CompositionSmtWriter<'a> {
...
         sample_id: &Option<usize>,
+        sample_name: &Option<String>,
         tipe: &Type,
     ) -> SmtExpr {
         let sample_id = sample_id.expect("found a None sample_id");
+        let sample_identifier = if let Some(sname) = sample_name {
+            sname
+        } else {
+            &format!("{}", sample_id)
+        };
```
keks, line 656: let's .expect() that this is set and pull the sample position out of the sample info (which should be in self or oracle_ctx), using the sample_id. debug assert that sample_name matches what is in the sample position.

src/writers/smt/writer.rs
```diff
@@ -658,8 +671,15 @@ impl<'a> CompositionSmtWriter<'a> {
...
+        let sample_id_struct: SmtExpr = (
+            "sample-id",
+            format!("\"{}\"", oracle_ctx.pkg_inst_ctx().pkg_inst_name()),
+            format!("\"{}\"", oracle_ctx.oracle_name()),
+            format!("\"{}\"", sample_identifier),
+        )
+            .into();
 
-        let rand_val: SmtExpr = (rand_fn_name, format!("{sample_id}"), ctr.clone()).into();
+        let rand_val: SmtExpr = (rand_fn_name, sample_id_struct, ctr.clone()).into();
```
keks, line 682: instead of defining the sample_id_struct, we can just use the sample position.

keks (changes requested):

keks (approved):
I think this looks good now!

PR #86, opened 2025-05-14
Fix/fun names in funcalls

### LINKED ISSUE #80: The way oracles call functions is wrong
Right now, if an oracle calls a function, it just calls some globally defined function. However, the name of that parameter is not in the name of the oracle function. This means that if we have two package instances of the same package that only differ in a function parameter, then either (a) the two definition collide or (b) only one of the two oracle functions is written, and the oracle code is wrong half the time.

Possible approaches:

1. not only put integer parameters into the names of functions, but also functions. This way we monomorphise wrt. the function assigments, which leads to flatter definitions but more SMT code.
2. we put function parameters into the proof const, game const and package const types. This means we can reuse the oracle code, but put functions into a datatype. This is not valid SMT-LIB 2.6, but should be supported by the solver we support. However, we don't know how well they support it.
3. pass the name of game instance and package instance into the oracle and decide which function to call based on that

Let's figure out which approach makes most sense to implement.

Possible Metrics:
- solver support/performance (solver may accept the code of solution 2, but can they still handle it?)
- smt code size
- smt code readability/navigatability
- implementation complexity

keks:
This PR changes the SMT-LIB code generation in such a way that functions are referred to by the name as declared in the proof. That way, things should be consistent. In order to do this, we do three things:

1. We properly populate the identifiers in FnCall with instantiation info. This way, a PackageConstIdentifier contains the value assigned in the game, and a GameConstIdentifier contains the value assigned in the proof. This is done recursively.
2. When declaring the functions that are declared in the proof, the name is changed to `<<func-{name_in_proof}>>`.
3. Whenever such a function is called, the name that is used in the proof is looked up by traversing the identifiers, and used to generate the correct name.

Fixes #80


src/identifier.rs
```diff
@@ -80,27 +100,19 @@ impl Identifier {
...
+            None
         }
     }
 
-    pub(crate) fn resolve_value(&self) -> Option<Expression> {
-        match self.value() {
-            Some(Expression::Identifier(ident)) => ident.resolve_value(),
-            Some(other) => Some(other.clone()),
-            None => None,
+    pub fn as_game_identifier(&self) -> Option<&game_ident::GameIdentifier> {
```
siccegge, line 111: This doesn't do recursive resolution which as_proof_identifier() does?

src/writers/smt/names.rs
```diff
@@ -41,7 +25,3 @@ pub(crate) fn fn_sample_rand_name<S: Into<SmtExpr>>(game_name: &str, rand_sort:
 pub(crate) fn oracle_nonsplit_arg_name(oracle_name: &str, arg_name: &str) -> String {
     format!("<arg-{oracle_name}-{arg_name}>")
 }
-
```
siccegge, line 44: also looks like a lot of cleanup in addition to actual changes? :grin:

siccegge:
looks reasonably sane can take another look when properly awake

src/identifier.rs
keks: No, and also I didn't actually use it so I got rid of it again

src/writers/smt/names.rs
keks: Yup :grin:

PR #89, opened 2025-05-23
Fix the way types (esp. Bits) are rewritten in oracle signatures for reductions

keks:


src/packageinstance.rs
```diff
@@ -369,19 +381,36 @@ pub(crate) mod instantiate {
             type_rewrite_rules
         }
 
+        pub(crate) fn rewrite_type(&self, ty: Type) -> Type {
```
siccegge, line 384: this only changes countspecs correct?

PR #98, opened 2025-06-06
Typecheck function calls

### LINKED ISSUE #96: Function Calls appear to not be typechecked in oracle code
In the Rosenpass code we had the encaps function that took the two arguments as a single tuple value, but then we called it with two non-tuple arguments. The error was only caught by the solver, but the typechecker should have caught this.

keks:
Fixes #96


siccegge:
like the use of expected_type :-)

PR #99, opened 2025-06-11
Fix the parameters included in the names of functions and datatypes in SMTLIB code

### LINKED ISSUE #79: Don't include boolean constant values in Game State names
We need to include integer constant names in the name, because the Bits types might be diffferent. However, right now we also include boolean values. This means we can't compare the states of two game instances that only differ in the value of a boolen constant parameter. While this is not wrong, it's a restriction that is not necessary.

keks:
Closes #79

### DIFF
--- a/src/writers/smt/names.rs
+++ b/src/writers/smt/names.rs
@@ -57,7 +57,7 @@ impl NameBuilderStage for NotEmpty {
 
 /// The core NameBuilder. Has a String that we append to when stuff is added.
 pub(crate) struct NameBuilder<Delim, Stage: NameBuilderStage> {
-    sort_name: String,
+    name: String,
     _delim: PhantomData<(Delim, Stage)>,
 }
 
@@ -76,7 +76,7 @@ impl<Delim: Delimiter> NameBuilder<Delim, Empty> {
         sort_name.push_str("<");
 
         Self {
-            sort_name,
+            name: sort_name,
             _delim: PhantomData,
         }
     }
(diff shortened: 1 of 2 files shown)


keks:
When reviewing, maybe start with src/writers/smt/patterns/instance_names.rs because that defines all the new types and traits.

src/writers/smt/patterns/datastructures/normal_return.rs
```diff
@@ -33,53 +34,52 @@ impl<'a> DatastructurePattern<'a> for ReturnPattern<'a> {
...
-            pkg_params,
-        } = self;
-
-        let game_encoded_params = encode_params(only_non_function_expression(*game_params));
-        let pkg_encoded_params = encode_params(only_non_function_expression(*pkg_params));
-        format!("<{camel_case}-{game_name}-{game_encoded_params}-{pkg_name}-{pkg_encoded_params}-{oracle_name}>")
+        let game_encoded_params = encode_params(only_ints(self.game_params));
+        let pkg_encoded_params = encode_params(only_ints(self.pkg_params));
+
+        SortNameBuilder::new()
```
siccegge, line 40: that also changes the output, though seems intentional/correct?

keks: Do you mean because the params are only added optionally? Yes, that is intentional.

siccegge: its using a SortName now and used FunctionName syntax before (which arguably was wrong)

PR #101, opened 2025-06-25
Fix randomness mapping when mapping makes claims about invalid IDs

siccegge:
When the randomness mapping makes claims about invalid sample ids (like if id left and right is equal), the current implementation will have an unsatisfiable mapping.

Change rand-is-eq to accept invalid sample ids instead of rejecting


keks (approved):
Nice, thank you two for finding and fixing it!

PR #112, opened 2025-07-16
Fix/103 108 encode consts in types to proof level

### LINKED ISSUE #103: Parameters encoded in names should always be resolved up to the proof level
When addressing #79 in #99 we found that when including parameters in names, we sometimes don't use the proof-level name but some intermediate one. This is most likely wrong.

### LINKED ISSUE #108: Bits(n) does not get along with randomness mapping
When building the `rand-is-eq` function, we look at the intersection between left and right types.
This is done somewhat simplistically and even if both left and right game inherit the same `n` from the proof their sampling types are different:

```
Bits(Identifier(PackageIdentifier(Const(PackageConstIdentifier { pkg_name: "Rand", name: "n", tipe: Integer, game_assignment: Some(Identifier(GameIdentifier(Const(GameConstIdentifier { game_name: "SmallComposition", name: "n", tipe: Integer, game_inst_name: None, proof_name: None, inst_info: None, assigned_value: None })))), pkg_inst_name: Some("rand"), game_name: Some("SmallComposition"), game_inst_name: None, proof_name: None }))))
```

vs

```
Bits(Identifier(PackageIdentifier(Const(PackageConstIdentifier { pkg_name: "Rand", name: "n", tipe: Integer, game_assignment: Some(Identifier(GameIdentifier(Const(GameConstIdentifier { game_name: "MediumComposition", name: "n", tipe: Integer, game_inst_name: None, proof_name: None, inst_info: None, assigned_value: None })))), pkg_inst_name: Some("rand"), game_name: Some("MediumComposition"), game_inst_name: None, proof_name: None }))))
```

Which results in an empty `rand-is-eq`

```
(define-fun rand-is-eq ((sample-id-left Int)
 (sample-id-right Int)
 (sample-ctr-left Int)
 (sample-ctr-right Int)
)
 Bool true)
```

keks:
This is still very unfinished, but it addresses two related issues:

- #103 
- #108 

so far I've only started looking at #108.

Closes #103
Closes #108

### DIFF
--- a/src/gamehops/equivalence/mod.rs
+++ b/src/gamehops/equivalence/mod.rs
@@ -1667,12 +1667,7 @@ impl<'a> EquivalenceContext<'a> {
                 .map(type_use_proof_ident),
         );
 
-        println!("randeq: types found in left:  {left_types:?}");
-        println!("randeq: types found in right: {right_types:?}");
-
         let types: Vec<&Type> = left_types.intersection(&right_types).collect();
-
-        println!("randeq: intersection: {types:?}");
 
         let mut left_positions_by_type: HashMap<_, Vec<_>> = HashMap::new();
         let mut right_positions_by_type: HashMap<_, Vec<_>> = HashMap::new();
--- a/src/gamehops/equivalence/mod.rs
+++ b/src/gamehops/equivalence/mod.rs
@@ -1632,6 +1632,23 @@ impl<'a> EquivalenceContext<'a> {
          *
          */
 
+        fn type_use_proof_ident(ty: Type) -> Type {
+            match ty {
+                Type::Bits(mut count_spec) => {
+                    if let CountSpec::Identifier(identifier) = count_spec.as_mut() {
+                        let proof_ident = identifier.as_proof_identifier();
+                        assert!(
+                            proof_ident.is_some(),
+                            "expected {identifier:?} to be completely resolved"
+                        );
+                        *identifier = Identifier::ProofIdentifier(proof_ident.cloned().unwrap());
+                    }
+                    Type::Bits(count_spec)
+                }
+                _ => ty,
+            }
+        }
+
         let left_positions = &self.sample_info_left().positions;
         let right_positions = &self.sample_info_right().positions;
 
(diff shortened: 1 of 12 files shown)


src/gamehops/equivalence/mod.rs
```diff
@@ -1631,29 +1632,65 @@ impl<'a> EquivalenceContext<'a> {
...
+        let right_types: HashSet<Type> = HashSet::from_iter(
+            self.sample_info_right()
+                .tipes
+                .iter()
+                .cloned()
+                .map(type_use_proof_ident),
+        );
+
+        println!("randeq: types found in left:  {left_types:?}");
+        println!("randeq: types found in right: {right_types:?}");
```
keks, line 1671: rm

src/gamehops/equivalence/mod.rs
```diff
@@ -1631,29 +1632,65 @@ impl<'a> EquivalenceContext<'a> {
...
 
         let types: Vec<&Type> = left_types.intersection(&right_types).collect();
 
+        println!("randeq: intersection: {types:?}");
+
```
keks, line 1676: rm

src/transforms/type_extract.rs
```diff
@@ -26,40 +31,52 @@ impl super::Transformation for Transformation<'_> {
     }
 }
 
+fn record_type(set: &mut HashSet<Type>, ty: Type) {
+    if let Type::Bits(cs) = &ty {
+        if let CountSpec::Identifier(ident) = cs.as_ref() {
+            println!(
+                "type extract: found Bits ident {:?}",
+                ident.as_proof_identifier()
+            )
```
keks, line 40: rm

src/transforms/type_extract.rs
```diff
@@ -14,8 +14,13 @@ impl super::Transformation for Transformation<'_> {
     fn transform(&self) -> Result<(Composition, HashSet<Type>), Infallible> {
         let mut set = HashSet::new();
 
+        // TODO: extract types of game state, params, oracle args, oracle return
+
         let insts = &self.0.pkgs.iter();
         let oracles = insts.clone().flat_map(|inst| inst.pkg.oracles.clone());
+
+        // TODO: extract types of package state, params, oracle args, oracle return
+
```
keks, line 23: are these todos still relevant?

PR #114, opened 2025-07-28
New UI and parallelism

### LINKED ISSUE #109: Parallelize Lemma Solving
Run multiple instances of the SMT solver in parallel.

### LINKED ISSUE #110: Make UI ready to report progress on multiple tasks in parallel
Maybe use [digestif](https://docs.rs/indicatif/latest/indicatif/)

siccegge:


src/ui/mod.rs
```diff
@@ -0,0 +1,237 @@
+use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
+
+pub(crate) struct ProofUI<'a> {
+    main_progress: MultiProgress,
+    current_proof: Option<&'a String>,
+    proofs_progress: Option<ProgressBar>,
+    seq_proof_progress: Vec<(String, ProgressBar)>,
+    seq_proofstep_progress: Vec<(String, String, ProgressBar)>,
+    seq_oracle_progress: Vec<(String, String, String, ProgressBar)>,
```
keks, line 9: I think these should probably be hashmaps

keks: Maybe find better names for these as well

keks (changes requested):

src/ui/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+
+pub(crate) struct ProofUI {
+    main_progress: MultiProgress,
+    project_progress: Option<ProgressBar>,
+    seq_proof_progress: HashMap<String, ProgressBar>,
+    seq_proofstep_progress: HashMap<(String, String), ProgressBar>,
+    seq_oracle_progress: HashMap<(String, String, String), ProgressBar>,
+}
+
+#[automock]
```
keks, line 13: I think I have two thoughts about this:

1. I would like to keep all these dependencies dev-dependencies. It seems like that is not possible with double, not sure. I don't know how that would work with automock.
2. Keep the magic low. I honestly am not sure how this works. double replaces the struct definition and automock replaces the impl? Don't know.

Maybe we could define a trait for the UI (with the functions basically being the block below) and then implement that once for the normal UI type an then inside test-only code define a second implementation using the [mock! macro](https://docs.rs/mockall/latest/mockall/macro.mock.html). The verify functions et al could then be made generic over that trait. Is that too much work? We could move that to later.

keks (commented):
I think this looks pretty good! I have two comments.

src/ui/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+
+impl ProofUI for IndicatifProofUI {
+    fn new(num_proofs: u64) -> Self {
+        let main_progress = MultiProgress::new();
+        let project_progress = if num_proofs > 1 {
+            let project_progress = main_progress.add(ProgressBar::new(num_proofs));
+
+            project_progress.set_style(indicatif_style::proof_bar());
+            project_progress.set_message("Project");
+            //project_progress.enable_steady_tick(std::time::Duration::from_secs(10));
```
keks, line 68: Is there a reason this is commented out?

src/ui/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+    );
+}
+
+pub(crate) struct IndicatifProofUI {
+    main_progress: MultiProgress,
+    project_progress: Option<ProgressBar>,
+    seq_proof_progress: HashMap<String, ProgressBar>,
+    seq_proofstep_progress: HashMap<(String, String), ProgressBar>,
+    seq_oracle_progress: HashMap<(String, String, String), ProgressBar>,
+}
```
keks, line 58: Could we move this, the impl and the style module into a submodule indicatif.rs?

keks (changes requested):
Just some nitpicks, I think we can get this in soon!

siccegge:
it has been done

keks (approved):
looks good!

PR #115, opened 2025-07-29
Make GameHop::Reduction Send:able by removing ast elements

siccegge:


keks:
I would prefer if we didn't lose the ability to access spans. I took the liberty and implemented the version I described in the chat on branch [feat/send-reductions](https://github.com/sspverif/sspverif/compare/feat/send-reductions).

siccegge:
Also works with the parallel stuff!

PR #119, opened 2025-08-06
set up env_logger and replace println!s by log messages

siccegge:
this sets up a very unconfigured env_logger.

set `RUST_LOG=debug` to see all println messages that were there before. This also very compatible with #114 via `indicatif_log_bridge`


keks (approved):
lgtm!

PR #123, opened 2025-08-20
Rewrite types in None and Sample expressions

keks:


src/packageinstance.rs
```diff
@@ -590,6 +590,8 @@ pub(crate) mod instantiate {
                 (_, Expression::FnCall(ident, args)) => {
                     Expression::FnCall(self.rewrite_identifier(ident), args)
                 }
+                (_, Expression::None(ty)) => Expression::None(self.rewrite_type(ty)),
+                (_, Expression::Sample(ty)) => Expression::Sample(self.rewrite_type(ty)),
```
siccegge, line 595: why do we have a `(_, ...` in every case instead of just matching expr?

siccegge (approved):
Clearly improving over previous work

PR #139, opened 2025-09-06
Conjectures

siccegge:


keks (approved):
I think this _should_ work, but maybe we can add an example and run it on CI?

PR #141, opened 2025-09-09
Feat/theorem tex

siccegge:
waiting for #128 first


src/writers/tex/mod.rs
```diff
@@ -1 +1,6 @@
+pub mod error;
```
keks: this look empty

keks (commented):
i'll look at the rest later!

src/parser/reduction.rs
```diff
@@ -1062,6 +1062,7 @@ pub(crate) struct ReductionMappingEntry<'a> {
 }
 
 impl<'a> ReductionMappingEntry<'a> {
+    #[allow(dead_code)]
     pub(crate) fn assumption(&self) -> &PackageInstanceName<'a> {
         &self.assumption
     }
```
keks: Maybe let's just remove the getter? We can trivially re-add it if we need it at some point.

keks (commented):
A bit more feedback, but it's difficult to review with all the moved code. maybe we can go through the rest together?

keks (approved):
Let's get this in - the tex export is best effort anyways

Issue #186, opened 2025-10-10
Integers in Identifier names


chrisbrzuska:
Cryptographers like to have names such as (msg_0,msg_1). Can we allow this in identifiers?

PR #187, opened 2025-10-15
all-zero / all-one constants for Bits_n (cf. #186)

siccegge:


siccegge:
really should not be allowed for Bits^*

keks (approved):
Looks good I think!

Issue #179, opened 2025-10-06
allow continuing incremental solving after retrieving a model


siccegge:
I would like to continue incrementally solving in the graph layouter even after I have extracted a model! should speed up solving significantly (almost all the constraints stay the same, just under tighter regimes wrt width/height of the graph).

However our code will close the connection to the solver after a get-model

PR #188, opened 2025-10-15
read_model only reads one model and keeps prover process open (cf. #179)

siccegge:


keks:
This will probably choke on strings that contain parentheses. We already talked about parsing SMTLIB style s-expressions, maybe this would be another application of that.

siccegge:
@keks pest only allows parsing "a &str". what would our strategy be? read more from the solver until the output parses and extract the remaining text from pest and put it in the buffer?

feels a bit unclean because we're re-starting from scratch trying to parse the output potentially many times but should work?

PR #204, opened 2025-11-28
Report ArgumentCountMismatch separately from TypeMismatch for functions (Closes: #203)

### LINKED ISSUE #203: Improve error message for typechecking function calls: wrong number of arguments
Currently, Domino only reports a type mismatch and does not say that the type mismatched is caused by a wrong number of arguments. Additionally, Domino reports the type of arguments as Unknown, Unknown, ... , Unknown. As a result, this error message looks like a Domino bug, but is actually proper type checking.

Suggestion: Improve error message so that it directly sais that the error is caused by a wrong number of arguments.

siccegge:


keks (approved):
nice!

PR #207, opened 2025-12-04
Use toplogical sort when writing packages to smt (Closes #75)

### LINKED ISSUE #75: Package Instances in Game Definitions are not Topologically Sorted
The code for packages is written in the order that they show up in package instance declarations in the games. This means that right how, we need to declare the package instances in such an order that for any package instance, all the oracle calls that it makes are satisfied. This should not be necessary, and we should either

- (a) do topological sorting of the package instances or
- (b) first just declare all the functions and then later constrain them to the specific behavior. This way, all identifiers are already defined by the time we talk about the actual content of the functions.

siccegge:


keks (approved):
looks good!

PR #210, opened 2025-12-04
 Check for matching exports in equivalence verification (closes 178)

### LINKED ISSUE #178: Oracle signatures of an equivalence gamehops are not checked to be matching
Domino does not check whether the signatures of the oracles exposed by the left and right games to be proved equivalent have matching signatures. As a result, the user only notices the error when the SMT code is generated and SMT solver returns a parse error that some arguments are not declared.

siccegge:


siccegge:
something is very much strange here (the $n$ in the Bits(n) types of helloworld have no proof assigned and thus matching the oracle signature fails)

tornado80 (approved):

keks (approved):
looks good!

let's create a followup though:

let's make the errors in gamehops::equivalence::error more like the one in the parser, where we have several individual structs that implement the miette types. This probably requires that we keep the spans around, though. Alternatively, move the check that returns these variants into the parser.

PR #214, opened 2025-12-09
pretty-printing smt

### LINKED ISSUE #205: UX: SMT pretty-printing?
Do you know of any tool that can pretty-print SMT nicely? That'd definitely help quite a lot when debugging SMT invariants, etc.
If we don't know of anything pre-existing, can you estimate how hard it would be to write? Feels doable, but maybe there are dragons...

Something like this maybe? https://github.com/symflower/smtfmt
Or can we just use cvc5 or z3 directly to parse and re-emit the assertions with nice formatting?
Discussed this with @tornado80

siccegge:


src/writers/smt/exprs.rs
```diff
@@ -218,25 +218,85 @@ impl<
...
+        ) -> std::fmt::Result {
+            match expr {
+                SmtExpr::Atom(str) => write!(f, "{str}"),
+                SmtExpr::List(lst) => {
+                    write!(f, "(")?;
+                    if lst.is_empty() {
+                    } else if matches!(lst[0], SmtExpr::List(_)) || lst.len() == 1 {
+                        let mut peek = lst.iter().peekable();
+
+                        while let Some(elem) = peek.next() {
```
luxas: I guess we don't need to peak the iterator, but can just use the iterator normally with a for loop?

luxas (commented):
Thanks a lot, looks great 🎉 
I could try out the pretty-printing of arbitrary-len fn applications within some bound if you want as well. 😄

keks (approved):
Looks good!

PR #221, opened 2026-01-05
Implement BaseUI, intialize logging early

### LINKED ISSUE #138: logging initialized too late
logging is only initialized when the UI is created (necessarily so!). However, as a result logging currently does not work in the parser!

siccegge:


src/ui/indicatif.rs
```diff
@@ -244,3 +224,46 @@ mod indicatif_style {
...
+            project_progress.set_style(indicatif_style::theorem_bar());
+            project_progress.set_message("Project");
+            Some(project_progress)
+        } else {
+            None
+        };
+        let logger = env_logger::Builder::from_default_env().build();
+        LogWrapper::new(main_progress.clone(), logger)
+            .try_init()
+            .unwrap();
```
keks:
```suggestion
```

siccegge:
This is a couple iterations on ui behind ;-)

PR #224, opened 2026-01-14
fix known-good example runner script

keks:


siccegge (approved):
Jep looks sane to me, good catch

PR #229, opened 2026-01-22
return an error when declaring variables twice

siccegge:
the following panicked and now returns an error:

```
theorem Test {
  const i: Integer;
  const i: Bool;

  gamehops {}
}
```


keks (approved):
Thanks!

PR #231, opened 2026-01-23
Support for Basic Hybrid Arguments

### LINKED ISSUE #228: Basic Hybrid Argument
maybe something like this?
```
theorem test {
    const n: Integer;
    const b: Bool;

instance Real = A{
  params {
    i: 0,
    b: false,
    ...    
  }
}​

instance Ideal = A{
  params {
    i: n,
    b: true,
    ...    
  }
}​

hybrid instance H i b = A {
  params {
    i: i,
    b: b,
    ...    
  }
}​


gamehops {
    hybrid H {
      reduction {
          assumption prf
          map Real_PRF H[b=false] {}
          map Ideal_PRF H[b=true] {}
      }
      equivalence {
          // H[i,b=true] = H[i+1,false]
          // proof obligation H:real:i
        ORACLE: {
          invariant [ ./theorem/invariant.smt2 ]
          lemmas {
            same-output: [no-abort]
            equal-abort: []    
            invariant: []
            }
          }
      }
    }

    equivalence Ideal B {}
    reduction B C {}
    equivalence C D {}
}

proposition {
    Real ~ D
}
}
```

siccegge:


keks (approved):
maybe remove the commented out code and rename the package instances (`b` is conventionally reserved for bits), otherwise looks good for now.

We need to figure out how to better write this parser code, but this is blocked on getting the spans into the AST

PR #252, opened 2026-01-30
Allow hybrids with different real/ideal games

siccegge:


src/parser/theorem.rs
```diff
@@ -386,18 +388,51 @@ fn handle_hybrid_instance_decl<'a>(
...
+        GameInstance::new(
+            format!("{game_inst_name}${bitval}${nextval}"),
+            theorem_name.to_string(),
+            game,
+            types,
+            consts,
+        )
+    }
+
+    let bit_var_name = if single {
```
keks: maybe `bit_var_name` should be `Option<&str>`

keks (changes requested):
looks pretty good, just a few asks

keks (changes requested):
I like this a lot more! One question, and then the point about the `clone` variable names still stands :)

keks (approved):

PR #255, opened 2026-02-04
Nix package manager support

koraa:


.github/workflows/nix.yml
```diff
@@ -0,0 +1,24 @@
+name: Nix-based Quality Assurance & Package Deployments
```
siccegge: maybe keep it concise? Don't see much QA in there. Maybe "Nix Build"

koraa: This is a bike shed, but all right.

src/writers/smt/tests/mod.rs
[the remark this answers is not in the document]
koraa: These are warnings emitted by clippy. They are required to make the nix based clippy run work.

siccegge (changes requested):
see comments above

siccegge:
For reference:

 1. A CI that turns red on an empty `ssp.toml` is not going in
 2. A CI that turns red on *unstable* clippy does not go in (only guessing here but we run latest stable as part of regular CI and it reported none of the findings)
 3. There's a few legacy projects with other toml files. We may decide to drop these projects but as long as they stay in these handwritten (!) toml files stay in a format that is readable

The rest is minor (and no blocker)

keks:
Closing this because we merged #295 (which was based on this - thanks @koraa!)

PR #258, opened 2026-02-04
testcase to reproduce #92

### LINKED ISSUE #92: Test the check that oracle signatures that are wired have matching types
Also see #56

siccegge:


keks (approved):
Thanks, let's merge this!

PR #261, opened 2026-02-05
allow oracle invocations without any return value (closes #28)

### LINKED ISSUE #28: Calling oracles without outputs
Our pseudocode currently only allows for invoke in an assignment, but when the oracle does not produce output, that's kind of an annoying restriction. Can we call such an oracle without an assignment?

siccegge:


src/parser/ssp.pest
```diff
@@ -188,7 +188,10 @@ ite = {"if" ~ expression ~ "{" ~ code ~ "}" ~ ( "else" ~ "{" ~ code ~ "}" )? }
 
 sample = { identifier ~ "<-$" ~ tipe ~ (kw_sample_name ~ identifier)? ~ ";" }
 assign = { identifier ~ "<-" ~ expression ~ ";" }
-invocation = { identifier ~ ( "[" ~ table_index ~ "]" )? ~ "<-" ~ kw_invoke ~ oracle_call ~ ";" }
+invocation = _{ invocation_return | invocation_noreturn }
```
keks:
```suggestion
invocation = _{ invocation_return | invocation_table | invocation_noreturn }
```

src/parser/package.rs
```diff
@@ -1123,20 +1124,30 @@ pub fn handle_code(
...
+                            let target_ident_name_ast = inner.next().unwrap();
+                            if target_ident_name_ast.as_str() == "_" {
+                                (inner, None, None)
+                            } else {
+                                (inner, Some(target_ident_name_ast), None)
+                            }
+                        } else {
+                            debug_assert_matches!(stmt.as_rule(), Rule::invocation_noreturn);
+                            (stmt.into_inner(), None, None)
+                        };
```
keks: This should probably be a match block

src/parser/package.rs
```diff
@@ -1123,20 +1124,30 @@ pub fn handle_code(
...
-                        (Some(index), inner.next().unwrap())
-                    } else {
-                        (None, maybe_index)
-                    };
+                Rule::invocation_table | Rule::invocation_return | Rule::invocation_noreturn => {
+                    let (mut inner, target_ident_name_ast, opt_idx) =
+                        if matches!(stmt.as_rule(), Rule::invocation_table) {
+                            let mut inner = stmt.into_inner();
+                            let target_ident_name_ast = inner.next().unwrap();
+                            assert!(target_ident_name_ast.as_str() != "_", "Special value _ disallowed for tables");
```
keks: we can only assert this if it is forbidden in the grammar, otherwise we need to return a proper error. Not sure that is the case.

keks (changes requested):
It would be cool to have this in! There are few things that need fixing first. The other question is whether we might want to first get #265 in. This will have some light conflicts, I suppose.

src/parser/package.rs
siccegge: we used to unwrap later on this :-D

Also there's the question of whether we want to keep `_` as special case? If so, should it be allowed everywhere? is our language side-effect-free enough that there's only reason to assign to `_` as part of an oracle call?

keks (approved):
I think this is good to merge. Let's get some tests in soon, though. Even an repro example project would be nice

PR #278, opened 2026-02-21
Rewrite User-Supplied SMT to support domino-specific macros

siccegge:


keks (changes requested):
i am super jetlagged and can't concentrate. this is what i found, otherwise i think we can merge this for now

PR #279, opened 2026-02-23
Make the Project a Trait in preparation for WebDomino

siccegge:
The short-term plan is to add a second project type that consumes a zip file instead of a folder, as in the current DominoWeb demo. Slightler longer perspective is a version that interactively loads and updates file contents from a web editor (possibly).

The command line implementation for zipfile is mostly to verify lifetimes do work out and one can satisfy the borrow checker.

Note: storing the bytes for the zipfile is a bit awkward, but reading from the zipfile needs *mutable* access (the implementation seeks inside the content) and we can't legally hand around mutable references to project so we construct a fresh cursor with access to the byte vec every time we need to access the file


crates/domino/src/main.rs
```diff
@@ -185,18 +240,20 @@ fn format(f: &Format) -> Result<(), project::error::Error> {
...
+fn wire_check(_game_name: &str, _dst_idx: usize) -> Result<(), project::error::Error> {
+    /*
     let project_root = project::find_project_root()?;
     let files = project::Files::load(&project_root)?;
     let project = project::Project::load(&files)?;
 
     project.print_wire_check_smt(game_name, dst_idx);
+    */
     Ok(())
 }
```
keks: should we just delete this?

crates/domino/src/main.rs
keks, line 1: should this maybe just have a completely separate main file?

src/project/mod.rs
```diff
@@ -7,164 +7,51 @@
...
 
-    pub fn load(files: &'a Files) -> Result<Project<'a>> {
-        let root_dir = find_project_root()?;
+pub mod error;
 
-        let packages: HashMap<_, _> = files
-            .packages()
-            .map(|pkg| pkg.map(|pkg| (pkg.name.clone(), pkg)))
-            .collect::<Result<_>>()?;
+pub trait Project: Sync {
```
keks: not sure this should always be Sync

src/project/mod.rs
```diff
@@ -7,164 +7,51 @@
...
+    fn get_root_dir(&self) -> PathBuf;
 
-        /* we already typecheck during parsing, and the typecheck transform uses a bunch of deprecated
-           stuff, so we just comment it out.
+    fn theorems(&self) -> impl Iterator<Item = &String>;
```
keks: should we return `&str` here instead of `&String`? Also, are thse the names that we can use below for getting the theorem (or game, respectively)? is there a reason we don't have this for packages?

keks (changes requested):
first round of review before my eyes fall shut :sleeping:

src/project/mod.rs
siccegge: because we don't have packages in the project ;-)
for some reason you only get those from the game

crates/domino/src/main.rs
siccegge: or we could just not have it.

I added this bit because it's nice to test the Zipfile variant of the project trait somewhere

I added the zipfile variant of the project trait because I need it for web anyways and it's nice to have several impls for the trait

but not sure having commandline support for the zipfile variant is all that useful

what do you think?

src/project/mod.rs
siccegge: but maybe we should have packages. would make my job on the web also easier!

siccegge: @keks what's the advantage of using `&str` here? I can change from `self.games.keys()` to `self.games.keys().map(|s| s.as_str())` but that seems almost a bit wasteful

siccegge: exposing packages now as well in the same way. should also be super helpful for web (-:

keks: There should not be waste, this is mapping over an iterator. And we can even do a point-free map with String::as_str :)

keks (approved):
Yess let's get this in!

PR #282, opened 2026-02-25
Make SmtSolver a trait

siccegge:
this allows to later have a cvc5 linked in directly or use some web magic

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,10 +25,13 @@ debug = true
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+
+[features]
+default = ["process-solver"]
+process-solver = ["dep:clap", "dep:expectrl", "dep:subprocess"]
+
 [dependencies]
-clap = { version = "4.0", features = ["derive"] }
 env_logger = "0.11"
-expectrl = "0.8.0"
 indicatif = "0.18"
 indicatif-log-bridge = "0.2"
 # Nice to have for further writer experiments:
@@ -42,10 +45,14 @@ rayon = "1.10"
 serde = "1.0.137"
 serde_derive = "1.0.137"
 sspverif-smtlib = { workspace = true }
-subprocess = "0.2"
 tempfile = "3"
 thiserror = "1.0.61"
 walkdir = "2"
+
+# Feature: process solver
+subprocess = { version = "0.2", optional = true }
+expectrl = { version = "0.8.0", optional = true }
+clap = { version = "4.0", features = ["derive"], optional = true }
 
 [dev-dependencies]
 mockall = "0.13"
--- a/src/gamehops/equivalence/verify_fn.rs
+++ b/src/gamehops/equivalence/verify_fn.rs
@@ -16,7 +16,7 @@ use crate::{
     project::Project,
     theorem::Theorem,
     transforms::{theorem_transforms::EquivalenceTransform, TheoremTransform},
-    util::prover_process::{Communicator, ProverBackend, ProverResponse},
+    util::smtsolver::{SmtSolver, SmtSolverBackend, SmtSolverResponse},
     writers::smt::exprs::SmtExpr,
 };
 
@@ -26,27 +26,29 @@ fn verify_oracle<UI: TheoremUI>(
     project: &Project,
     ui: Arc<Mutex<&mut UI>>,
     eqctx: &EquivalenceContext,
-    backend: ProverBackend,
+    backend: &impl SmtSolverBackend,
     transcript: bool,
     req_oracles: &[&Export],
 ) -> Result<()> {
     let eq = eqctx.equivalence();
     let proofstep_name = format!("{} == {}", eq.left_name(), eq.right_name());
 
-    let mut prover = if transcript {
+    let mut solver = {
         let oracle = if req_oracles.len() == 1 {
             Some(req_oracles[0].name())
         } else {
             None
         };
 
-        let transcript_file: std::fs::File = project
-            .get_joined_smt_file(eq.left_name(), eq.right_name(), oracle)
-            .unwrap();
+        if transcript {
+            let transcript_file: std::fs::File = project
+                .get_joined_smt_file(eq.left_name(), eq.right_name(), oracle)
+                .unwrap();
 
-        Communicator::new_with_transcript(backend, transcript_file)?
-    } else {
-        Communicator::new(backend)?
+            backend.new_smtsolver_with_transcript(transcript_file)?
+        } else {
+            backend.new_smtsolver()?
+        }
     };
     std::thread::sleep(std::time::Duration::from_millis(20));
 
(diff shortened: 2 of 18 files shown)


src/util/smtsolver/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProverResponse {
+    Sat,
+    Unsat,
+    Unknown,
+}
+
+pub trait ProverFactory {
+    fn new_prover(&self) -> Result<impl Prover>;
+    fn new_prover_with_transcript<W: std::io::Write + Send + Sync + 'static>(
```
keks: `'static` is not nice, but maybe not a terrible limitation in this case :thinking:

src/util/smtsolver/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+pub use error::{Error, Result};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProverResponse {
+    Sat,
+    Unsat,
+    Unknown,
+}
+
+pub trait ProverFactory {
```
keks: maybe `ProverBackend`, `SolverBackend`, `SmtSolverBackend` or `SmtBackend`?

src/util/smtsolver/process.rs
```diff
@@ -21,63 +20,41 @@ pub enum ProverBackend {
...
-        matches!(self, Self::Unknown)
+    fn new_prover_with_transcript<W: std::io::Write + Send + Sync + 'static>(
+        &self,
+        writer: W,
+    ) -> Result<impl Prover> {
+        if self.transcript {
+            Communicator::new_with_transcript(self.backend, writer)
+        } else {
+            Communicator::new(self.backend)
+        }
```
keks: I think this means that if self.transcript is false, then we first open (create) a file and then don't write to it. I think that decision needs to happen earlier.

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -26,14 +26,14 @@ fn verify_oracle<UI: TheoremUI>(
...
-    transcript: bool,
+    backend: &impl ProverFactory,
+    _transcript: bool,
     req_oracles: &[&Export],
 ) -> Result<()> {
     let eq = eqctx.equivalence();
     let proofstep_name = format!("{} == {}", eq.left_name(), eq.right_name());
 
-    let mut prover = if transcript {
+    let mut prover = {
```
keks: Maybe just undo these, then we only create the file inside the branch.

src/util/smtsolver/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn new_prover(&self) -> Result<impl Prover>;
+    fn new_prover_with_transcript<W: std::io::Write + Send + Sync + 'static>(
+        &self,
+        write: W,
+    ) -> Result<impl Prover>;
+    fn transcript_enabled(&self) -> bool;
+}
+
+pub trait Prover: fmt::Write {
```
keks: Maybe call it `Solver` or `SmtSolver`?

keks: Also, does it need to impl Write? Maybe that makes things easier, not sure. Might make things more complicated with some backends (but maybe not a problem for now)

src/util/mod.rs
```diff
@@ -1,7 +1,9 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+#[cfg(feature = "process-solver")]
 pub mod process;
-pub mod prover_process;
+
+pub mod prover;
```
keks: Call it `smt_solver`?

src/util/smtsolver/mod.rs
keks: either put the prover type as a generic argument of the trait, or as an associated type:

```rust
pub trait ProverFactory<P: Prover> {
  fn new(..) -> Result<P, ...>;
```

or

```rust
pub trait ProverFactory {
  type Prover: Prover;
  fn new(..) -> Result<Self::Prover, ...>;
```

keks (changes requested):
first round of review

src/util/smtsolver/mod.rs
keks: The first allows implementing the trait several times, for different concrete prover types. probably not needed

src/util/smtsolver/mod.rs
siccegge: I think it's still used in some obscure corners but maybe a good point to get rid of it

siccegge: @keks actually, impl Write is quite convenient and used *everywhere*. and not sure having a write_str() would be niver?

src/util/smtsolver/mod.rs
siccegge: this comes from

```rust
impl Communicator {
    pub fn new_from_cmd<W: std::io::Write + Send + Sync + 'static>(...) {
        [...]
        let thrd = std::thread::spawn(move || {
            [...]
        };
        [...]
    }
}
```

I guess we can replace it with `Arc`. I suggest we do that separately though, should we open an issue?

src/util/smtsolver/mod.rs
siccegge: backend we already use for the enum (process can do cvc, z3). It could be `Prover<Cvc5>` though (also connects to discussion further down ;-) ) as the options in the enum will most likely be specific for each prover type.

src/util/smtsolver/mod.rs
keks: - #312

siccegge:
So. Should address everything :grin:

Cargo.toml
```diff
@@ -25,10 +25,13 @@ debug = true
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+
+[features]
+default = ["process-solver"]
```
keks, line 30: should this be a default feature or should this be in the domino dependency? can stay for now

src/project/mod.rs
```diff
@@ -220,7 +220,7 @@ impl<'a> Project<'a> {
     // we could then extract the theorem viewer output and other useful info trom the trace
     pub fn prove(
         &self,
-        backend: ProverBackend,
+        backend: &(impl SmtSolverBackend + Sync),
```
keks, line 223:
```suggestion
        backend: &impl SmtSolverBackend + Sync,
```

src/util/process.rs
```diff
@@ -45,7 +45,10 @@ impl Communicator {
         let stdout = cmd.stdout.unwrap();
 
         let thrd = std::thread::spawn(move || {
-            writeln!(stdin)?;
+            write!(stdin, "")?;
+            if let Some(ref mut transcript) = transcript {
+                write!(transcript, "")?;
+            }
```
keks, line 53: maybe flush to make sure this works also with line buffering?

keks (approved):
Looks great! These are mostly comments, not change requests. I think we can drop the parens around the impl, though.

PR #286, opened 2026-02-27
UI Rewrite

### LINKED ISSUE #197: Progress bars depend on terminal height
When the terminal is not sufficiently high to show all progress bars, then the showing of progress bars is buggy: 
- The top bar showing how many proof steps were successfully verified, disappears, when prover terminates.
- Progress bars for later steps might not be shown.

### LINKED ISSUE #246: LaTeX Export Progress bar
LaTeX Export could have a progress bar, too :-). Currently, it takes a long time and one does not see what is happening. In particular, it might be hard for a user to guess that in the back

siccegge:


siccegge:
Note: the indicatif structures are essentiallly just an alias for Arc<Mutex<Impl>>  so cloning them instead of passing references is fine (and avoids us struggling with parallel)

PR #287, opened 2026-03-01
shortform assign lists

siccegge:


keks (approved):
Let's get this in as-is and open a followup for doing something similar for the reduction mapping. We looked over this today and that seems to be the only remaining place where having something like this would make sense.

PR #291, opened 2026-03-17
More expressive --version (now with branch & commit!)

siccegge:
Bit unfortunate but we need to pull out the Command structure to build.rs/main.rs to avoid circular dependencies


keks (approved):
Looks good! Let's open a followup to discuss where shell completion generation should live in the long run.

PR #292, opened 2026-03-24
Small fixes for oracle renaming

siccegge:


keks (changes requested):
Let's add a test here

keks (approved):
Thanks!

PR #294, opened 2026-04-01
Clean Up

siccegge:


siccegge:
(the one file **deleted** is in `_build` and should have never been commited)

keks (approved):

PR #295, opened 2026-04-01
Add Nix Flake for running Domino

keks:
This is based on @koraa's PR, thank you very much! We have adapted it a little to simplify the control flow and do less formatting.

Before merge:
- [ ] credit koraa in commit msg
- [x] ensure CI works
- [x] add more architectures to CI (mac, arm linux, ~~32bit linux~~)
  - neither github actions nor nix seem to support 32bit very well, and it is not super likely to be used anyway

### DIFF
--- a/.github/workflows/known-good-proofs.yml
+++ b/.github/workflows/known-good-proofs.yml
@@ -2,7 +2,7 @@ name: Run Known-Good Proofs
 
 on:
   pull_request:
-    branches: [ main ]
+    branches: [main]
   workflow_dispatch:
 
 jobs:
@@ -30,4 +30,3 @@ jobs:
           name: Project PDFs
           path: ./example-projects/**/_build/latex/Theorem_*.pdf
           retention-days: 14
-
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,26 @@
+name: Test Nix Flake
+
+on:
+  pull_request: {}
+  push:
+    branches:
+      - main
+
+concurrency:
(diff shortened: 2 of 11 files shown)


koraa:
@keks LMK when you would like a review :)

keks:
@koraa Thanks, yeah I'd appreciate a review! The only think I see that is left to do here is to squash and attribute you in the commit message, which I can do through the Github Web UI after the review is done.

siccegge:
nb: this seems to still have the issue of triggering clippy warnings that rust/clippy/latest does not find

koraa (approved):
The PR looks good to me.

We had a discussion about coding style before; please feel free to disregard what follows.

Personally, I find the flake.nix extremely hard to read now; it is following standard nix coding style, which is a bad coding style in my opinion.

In terms of objective metrics, the file now has 6 levels of nesting, which is more than I can easily keep track of. If you also feel that this is an issue, you could improve the situation by moving the main piece of the code, i.e. the function passed to `flake-utils.lib.eachDefaultSystem` into a separate file (main.nix or nix/main.nix for instance). I would also recommend that you simplify the long `let <vars>; in <outputs>` statement by rewriting to `let <vars>; outputs = <outputs>; in outputs`. This way the actual body of the `let … in …` statement becomes very short and you can get most of the advantages of the `scoped` trick without directly relying on it as a device.

keks:
Thank you for your review @koraa! I took your note on the deep nesting and large body of the let expression and turned that into something that seems sensible to me. I also made a few variables more explicit. I'll merge this once CI is happy. Thanks for your input and for providing a base to work off of!

PR #297, opened 2026-04-09
Notify the user on (partially) empty projects (#296)

### LINKED ISSUE #296: Check for appropriate folders when invoking domino prove
Running `domino prove` or `RUST_BACKTRACE=1 domino prove` on code written for older domino versions (when the theorem folder was proof) does not generate an error or warning, and silently quits. An error message might be nice to have.

siccegge:


siccegge:
This is, of course, very hacky and very temporary. We'll still want something along the lines after #286 but for now it's helpful anyhow.

In detail:
 - we could do miette type errors and end execution. I do not want this -- current behaviour is useful
 - we could do log::warn!() but currently logging is not initialized when this code runs. This will change after #286 and we can do that then

keks (approved):

PR #298, opened 2026-04-15
make order of cases in rand-is-eq deterministic

siccegge:


siccegge:
as reported by @tornado80 :upside_down_face:

keks (approved):

PR #300, opened 2026-04-16
Fix Proof Assembly

### LINKED ISSUE #268: Proof assembly fails to deal with "splitting a bit"
in 4WHS/full we go from `H5[b]` to `H6[btest=b,bleast=b]`, for `H6` we have a step `H6[btest=false,bleast]` to `H6[btest=true,bleast]`  and can finally proceed from `H6[btest=true,bleast]` to `H7[bleast]`.

Proof assembly should find

```
H5[false]
->
H6[false,false]
->
H6[true,false]
->
H7[false]
->
H7[true]
->
H6[true,true]
->
H5[true]
```

but fails to properly construct the first step
```
thread 'main' (165847) panicked at src/proof.rs:246:26:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

siccegge:


keks (approved):
I think this probably looks good. I think we should come up with a more abstract description of this algorithm that is easier to wrap the head around than the Rust code, because I notice I keep having trouble wrapping my mind around the details here.

PR #301, opened 2026-04-16
Run check on windows as well

siccegge:


scripts/test-known-examples.sh
```diff
@@ -42,6 +42,7 @@ sed -e 's/[[:space:]]*#.*// ; /^[[:space:]]*$/d' "example-projects/known-good.tx
       cd _build/latex
       for file in Theorem_*.tex; do
         TEMPFILE=$(mktemp)
+        echo "### Runnin latex on $file"
```
keks:
```suggestion
        echo "### Running latex on $file"
```

src/util/smtparser/testdata/.gitattributes
```diff
@@ -0,0 +1 @@
+* -text
```
keks:
```suggestion
# prevent line ending conversion in this directory
* -text
```

keks (approved):
just 2 small things

PR #302, opened 2026-04-19
Generated Randomness Mappings

### LINKED ISSUE #162: re-usable randomness mapping
It would be nice if I can have only one (or two) randomness mapping functions.
Especially with sample-ids and "good usage" of instance names very often the trivial mapping

```lisp
(define-fun randomness-mapping-NewKey
    ((base-ctr-0 Int) (base-ctr-1 Int)
     (id-0 SampleId) (id-1 SampleId)
     (scr-0 Int) (scr-1 Int))
  Bool
  (and (= base-ctr-0 scr-0)
       (= base-ctr-1 scr-1)
       (= id-0 id-1)))
```

works for (almost) all oracles and repeating this function 14 times for each oracle is slightly annoying

siccegge:


src/parser/ssp.pest
```diff
@@ -64,6 +64,9 @@ kw_gamehops = _{ "gamehops" }
 kw_map = _{ "map" }
 kw_invariant = _{ "invariant" }
 kw_lemmas = _{ "lemmas" }
+kw_randomness = _{ "randomness" }
+kw_randomness_auto = { "auto" }
+kw_randomness_none = { "none" }
```
keks: let's drop these and use identifiers instead

src/gamehops/equivalence/mod.rs
```diff
@@ -783,6 +796,75 @@ impl<'a> EquivalenceContext<'a> {
         Ok(())
     }
 
+    fn emit_auto_randomness(&self, comm: &mut Communicator, oracle_name: &str) -> Result<()> {
+        match self.equivalence.randomness_by_oracle_name(oracle_name) {
+            RandomnessType::Custom => {}
+            RandomnessType::Auto => {
```
keks: Maybe we should call this Simple instead of Auto, because we may come up with more auto-mappings in the future? not sure

keks (changes requested):
generally looks good, but let's change these minor things

src/parser/theorem.rs
```diff
@@ -987,10 +987,10 @@ fn handle_equivalence_oracle(
...
         match next.as_rule() {
             Rule::randomness_spec => {
-                let value = next.into_inner().next().unwrap().as_rule();
+                let value = next.into_inner().next().unwrap().as_str();
                 match value {
-                    Rule::kw_randomness_auto => randomness = RandomnessType::Auto,
-                    Rule::kw_randomness_none => randomness = RandomnessType::None,
+                    "simple" => randomness = RandomnessType::Simple,
+                    "none" => randomness = RandomnessType::None,
                     _ => unreachable!(),
```
keks, line 994: unfortunately not anymore :)

keks (changes requested):

src/parser/theorem.rs
siccegge: too bad

siccegge: now we need to have an actually useful error message. on it

PR #307, opened 2026-04-28
Verify Completeness of Induction Step

siccegge:
In this PR we make sure that all the claims used as dependencies to show that the invariant in equivalence proofs actually holds are checked. So far the user had to manually ensure that they tell Domino to check all of them.

In order to not break some of our more complex examples where we do manual checks, we introduce the `admit` keyword that tells domino to not prove this claim.

- [x] add a warning when we encounter `admit`

### DIFF
--- a/src/gamehops/equivalence/verify_fn.rs
+++ b/src/gamehops/equivalence/verify_fn.rs
@@ -104,28 +104,30 @@ fn verify_oracle<UI: TheoremUI>(
                 claim.name(),
             );
 
-            writeln!(prover, "(push 1)").unwrap();
-            eqctx.emit_claim_assert(&mut prover, export.name(), &claim)?;
-            match prover.check_sat()? {
-                ProverResponse::Unsat => {}
-                response => {
-                    let modelfile = prover.get_model().map(|(modelstring, _model)| {
-                        let mut modelfile =
-                            tempfile::Builder::new().suffix(".smt2").tempfile().unwrap();
-                        modelfile.write_all(modelstring.as_bytes()).unwrap();
-                        let (_, fname) = modelfile.keep().unwrap();
-                        fname
-                    });
-                    prover.close();
-                    return Err(Error::ClaimTheoremFailed {
-                        claim_name: claim.name().to_string(),
-                        oracle_name: export.name().to_string(),
-                        response,
-                        modelfile,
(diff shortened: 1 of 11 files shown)


siccegge:
(depends on #304 as it also needs result versions of `handle_equivalence_oracle` and we don't want to conflict)

src/parser/theorem.rs
```diff
@@ -1063,28 +1135,64 @@ fn handle_equivalence_oracle(
             Rule::invariant_spec => {
                 invariant_paths.extend(handle_invariant_spec(next.into_inner()));
             }
-            Rule::lemmas_spec => lemmas.extend(handle_lemmas_spec(next.into_inner())),
+            Rule::lemmas_spec => {
+                let span = next.as_span();
+                let new_lemmas = handle_lemmas_spec(ctx, oracle_name, next.into_inner());
+                verify_induction_step(ctx, &new_lemmas, (span.start()..span.end()).into())?;
```
keks, line 1141: Does this have to be here? I guess, for nice error messages... I really need to finish that parser rewrite.

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -104,28 +104,30 @@ fn verify_oracle<UI: TheoremUI>(
...
-                    });
+            if !claim.is_admitted() {
+                writeln!(prover, "(push 1)").unwrap();
+                eqctx.emit_claim_assert(&mut prover, export.name(), &claim)?;
+                match prover.check_sat()? {
+                    ProverResponse::Unsat => {}
+                    response => {
+                        let modelfile = prover.get_model().map(|(modelstring, _model)| {
+                            let mut modelfile =
+                                tempfile::Builder::new().suffix(".smt2").tempfile().unwrap();
```
keks, line 115: I know this is existing code, but maybe we should open a followup for not using .smt2 here, because that sounds like it is an smt-lib source file (at least to me)

keks (approved):
Two remarks, but I think we can merge this now!

src/gamehops/equivalence/verify_fn.rs
siccegge: i mean it is. and you want smt2 syntax hilighting

PR #309, opened 2026-05-02
Report duplicate instance definitions (#308)

### LINKED ISSUE #308: Should warn on defining multiple game instances of the same name
```
theorem {
   instance foo = bar {}
   instance foo = baz {}
}
```
should produce a nice big warning!

siccegge:


keks:
I think this looks good, but I think a test would be nice.

PR #315, opened 2026-05-10
load invariant files relative to project root (#314)

### LINKED ISSUE #314: Domino invoked in a subdirectory of a project complains about location of invariant files
When Domino is invoked in a subdirectory of a project (e.g. theorems or packages), it can not find the root of the project and complains about missing invariants relative to the current working directory instead of project root.

siccegge:


keks:
This is also solved by the project trait refactor

PR #324, opened 2026-06-01
Nice error when parsing non-tuples

### LINKED ISSUE #323: Type indications for parsing tuples are off
The following code:

```
package pkg {
  oracle test() {
    x <- 12;
    (a, b) <- x;
  }
}
```

produces 

```
Error: domino::code::type_mismatch

  × type mismatch: got Integer, expected Unknown
   ╭─[pkg.pkg.ssp:4:5]
 3 │     x <- 12;
 4 │     (a, b) <- x;
   ·     ───┬──
   ·        ╰── this expression has the wrong type
 5 │   }
   ╰────
```

note the `expected Unknown` which should at the very least say `expected (Unknown, Unknown)` but probably it should tell us that we are trying to deconstruct a non-tupole type!

further if we modify the code to

```
package pkg {
  oracle test() {
    x <- 12;
    a <- 5;
    (a, b) <- x;
  }
}
```

we now *know* that the first value (`a`) should be an Integer!

siccegge:


keks (approved):
Good call, thanks!

PR #326, opened 2026-06-03
Move graph layouter out of writer::tex

siccegge:


keks (approved):
I wonder if src/writers/smt_graphlayout or so would have been better, but I think this is good too.

PR #327, opened 2026-06-03
implement caching for composition graphs

siccegge:


src/writers/util/graph.rs
```diff
@@ -102,7 +134,8 @@ pub(crate) fn solve_composition_graph(
 
     log::debug!("Conclusion: height = {max_height}, width = {max_width}");
 
-    model
+    let _ = write(path.join("model"), modelstring);
```
keks: This returns a Result that should be checked, no?

src/writers/util/graph.rs
```diff
@@ -30,27 +40,49 @@ impl<'comp> GraphLayout<'comp> {
...
+        if stored_constraints == constraints {
+            if let Ok(modelstring) = read_to_string(path.join(&composition.name).join("model")) {
+                log::info!("found graph for {} in cache!", &composition.name);
+                return SmtModel::from_string(&modelstring);
+            }
+        }
+    }
+    path.push(&composition.name);
+    std::fs::create_dir_all(&path).unwrap();
+    let _ = write(path.join("constraints"), &constraints);
```
keks: This also drops a result.

src/writers/util/graph.rs
```diff
@@ -30,27 +40,49 @@ impl<'comp> GraphLayout<'comp> {
...
+        log::debug!(
+            "found potential graph for {} in cache, checking for matching constraints",
+            &composition.name
+        );
+        if stored_constraints == constraints {
+            if let Ok(modelstring) = read_to_string(path.join(&composition.name).join("model")) {
+                log::info!("found graph for {} in cache!", &composition.name);
+                return SmtModel::from_string(&modelstring);
+            }
+        }
```
keks, line 72: Should we delete the cached values in an else case here? Sounds like we need to invalidate the cache.

keks (changes requested):
I think this drops two results that shouldn't be dropped. Clippy also complains (looks pretty easy to address).

I am wondering: Should we double-check that the model in the cache actually satisfies the contraints? That would allow us to recover from a somehow poisoned cache.

src/writers/util/graph.rs
siccegge: it's "just" a cache so maybe if we can't write it should not be fatal. but it should deinitely warn the user :upside_down_face:

src/writers/util/graph.rs
siccegge: We're about to overwrite both files in a second so that should take care of it!

keks (approved):
Looks good, I think!

PR #328, opened 2026-06-03
print types on oracle sig mismatch

siccegge:


src/package.rs
```diff
@@ -23,7 +26,31 @@ pub struct OracleSig {
...
+            f,
+            "({}) -> {}",
+            self.args.iter().map(|t| format!("{t}")).join(", "),
+            self.ret
+        )
+    }
+}
+
 impl OracleSig {
+    pub(crate) fn as_oracle_type(&self) -> OracleSigType {
```
keks: I think "as" is wrong here. Maybe to, or we make it consume self and make it into? https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

src/package.rs
```diff
@@ -23,7 +26,31 @@ pub struct OracleSig {
...
+pub struct OracleSigType {
+    args: Vec<Type>,
+    ret: Type,
+}
+
+impl Display for OracleSigType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(
+            f,
+            "({}) -> {}",
```
keks: should this be "oracle ({}) -> {}"? not sure...

keks (changes requested):
two small nits, otherwise great!

keks (approved):
Perfect, thanks!

PR #331, opened 2026-06-11
Package & Game Invariants

siccegge:

PR #332, opened 2026-06-14
remove incremental mode from domino

siccegge:


src/gamehops/equivalence/verify_fn.rs
```diff
@@ -32,58 +31,45 @@ fn verify_oracle<UI: TheoremUI>(
...
     log::debug!(
         "emitting const declarations for {}-{}",
         eq.left_name,
         eq.right_name
     );
-    eqctx.emit_constant_declarations(&mut solver)?;
+    smt.append(&mut eqctx.emit_constant_declarations());
 
     for export in req_oracles {
+        let mut smt = smt.clone();
```
keks: maybe let's make this a separate list and iterate over all our lists to populate the solver at the end

src/gamehops/equivalence/emit.rs
```diff
@@ -0,0 +1,961 @@
...
+            oracle_args::UnitOracleArgPattern, theorem_constants::ConstantPattern,
+            ReturnIsAbortConst, SmtDefineFun,
+        },
+        sorts::Sort,
+        writer::CompositionSmtWriter,
+    },
+};
+
+impl<'a> EquivalenceContext<'a> {
+    pub(super) fn emit_invariant(&self, oracle_name: &str) -> Result<Vec<SmtExpr>> {
```
keks: Can we prepare the rewritten invariants somehow? Then this function can be infallible

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -107,8 +92,30 @@ fn verify_oracle<UI: TheoremUI>(
             );
 
             if !claim.is_admitted() {
-                writeln!(solver, "(push 1)").unwrap();
-                eqctx.emit_claim_assert(&mut solver, export.name(), &claim)?;
+                let mut smt = smt.clone();
```
keks: same here

keks (commented):
I think I generally like the idea. A few suggestions that we discussed could be worthwhile

src/gamehops/equivalence/emit.rs
siccegge: actually this would combine very well with a concern I have in #331 -- I really want to combine package invariants from possibly multiple files into one statement -- which will be easy to do!

keks (approved):
I think this looks good. We definitely need better testing (incl. negative cases) of the equivalence proving part. But I agree the architectural changes are important, an we can clean up issues later, if there are any.

PR #337, opened 2026-06-24
Linebreaks in error messages

### LINKED ISSUE #335: Test for nice error messages
We temporarily lost nice error messages (see #334) and the tests did not trigger. we should capture this in our tests

siccegge:


keks (approved):
sure, why not

PR #338, opened 2026-06-24
Clean yao and kem-dem

tornado80:


siccegge:
Note: please consider adding the projects to `example-projects/known-good.txt`!

siccegge:
```
kem-dem-cca-security-easycrypt/easycrypt/Tutorial_KEMDEM.ec
[critical] [kem-dem-cca-security-easycrypt/easycrypt/Tutorial_KEMDEM.ec: line 519 (13)] parse error
[-] [0519] 42.0% (-1.0B / [frag -1.0B])
```

tornado80:
Apparently `EasyCrypt@r2026.05-23-gc64f7a3` and `Z3@4.13.4` work.

chrisbrzuska (approved):
I like the new structure of the different KEM-DEM projects a lot, especially the comparison with EasyCrypt also. For the SSP variant, we can consider adding a variant with reduction to correctness at some later point---but no need to think about this now. I think that it's good to prioritize having a tutorial first and then, we can expand it later.

For Yao, I hope that we can simplify later via "foreach" with primitive loop unrolling once we have decided whether we merge PR #348 .

PR #339, opened 2026-06-24
fix bit constants

siccegge:


keks (approved):
:face_with_peeking_eye:

keks:
let's add a test here

siccegge:
technically we run proofsteps on 4WHS so together with @chrisbrzuska 's change we should test for it now. but adding an explicit test :upside_down_face:

PR #340, opened 2026-06-24
Use pre-defined constants in 4WHS

chrisbrzuska:
We now use pre-defined constants in 4WHS.


siccegge (approved):
so much nicer without the parameter noise

PR #342, opened 2026-06-24
error help

siccegge:


src/parser/error.rs
```diff
@@ -33,7 +33,10 @@ pub struct AdmittedClaimWarning {
 
 #[derive(Error, Diagnostic, Debug)]
 #[error("could not prove claim {target}. provable claims: {provable}")]
-#[diagnostic(code(domino::code::theorem::inductionstep::unprovable))]
+#[diagnostic(
+    code(domino::code::theorem::inductionstep::unprovable),
+    help("claims are only provable if they are admited or all dependencies are provable")
```
keks:
```suggestion
    help("claims are only provable if they are admitted or all dependencies are provable")
```

keks (approved):
one type, looks good otherwise :+1:

PR #344, opened 2026-06-25
Fix SMT error if left and right exports use different *names* for ora…

siccegge:
…cle arguments


siccegge:
Problem: Old code assumes that the arguments left/right have the same names. This breaks if the user uses different names. It becomes a soundness bug if the user uses colliding names left/right with different positions/meaning

keks (approved):
Good catch, thanks!

PR #345, opened 2026-06-25
Small latex improvements

siccegge:


src/writers/tex/writer/util.rs
```diff
@@ -0,0 +1,60 @@
...
+        let mut rest: Vec<_> = it.map(|s| format!("{join}{s}")).collect();
+        lines.append(&mut rest);
+        format!(
+            "\\begin{{array}}{{c}}{}\\end{{array}}",
+            lines.join("\\pclb")
+        )
+    }
+}
+
+pub(super) fn list_to_matrix(mut it: impl Iterator<Item = String>) -> String {
```
keks: I think the idiomatic thing here would be to use an `IntoIterator<Item = &str>`. Then it should be possible to just pass in a slice of strings. Not sure if this makes lifetimes annoying - if that is the case, then we could try

```rust
fn list_to_matrix<S: Borrow<str>>(mut it: impl Iterator<Item = S>) -> String
```
which includes both String and &str

keks (changes requested):
I think this looks pretty good. I have one suggestion for making the code a little bit more "rusty" I guess, but it doesn't really matter here anyways.

One idea (not for this PR): You could have a type `Latex<T>(T)` and then implementations of `Display` for `Latex<Type>`, `Latex<Expression`, etc (actually maybe also `Latex<&Type>' etc).Then you can just wrap them in the constructor to make the thing printable.

Not sure that makes the code a lot better, just a thought I had when looking at this.

src/writers/tex/writer/util.rs
siccegge: @keks I have now `<S: ToString + Clone + Borrow<str>>` but it doesn't allow me to call len() on the element

PR #347, opened 2026-06-29
loop unrolling

siccegge:


keks (approved):
I always have trouble reasoning about code that uses expr.map et al., but I think this looks good!

PR #349, opened 2026-06-30
Allow offset-based randomness mapping (closes #181)

### LINKED ISSUE #181: Simplify randomness mapping relation
currently we have for both left and right the arguments for the counter at the time when the attacker calls the oracle and then the counter for when the randomness is sampled. In reality, all we need is the offset.

In addition, it seems like the solver is faster if we only use the offset.

So let's change the function to offset-only.

tornado80:
This change simplifies the randomness mapping relation such that the user only needs to provide the sampling id's and zero-based sampling offsets. The current approach is that we constrain the randomness counters in the old state to be zero and directly pass offsets to the `rand-is-eq` function.


keks (approved):
Perfect, thanks!

PR #355, opened 2026-07-13
Buffer resizing for reading from process

### LINKED ISSUE #353: Buffer not large enough for some models
in `src/util/process.rs`

```rust
        let buf = vec![0u8; 1 << 20];
```

which is not large enough for some of the models we care about. Changing to 16MiB makes my project work but maybe we need something smarter than a constant?

siccegge:
The concrete values (resize once below 10% and double the buffer size) are just yolo'd. in particular, doubling avoids messing with floating point conversion -- I guess one could do `*15/10` or so as well if we prefer slower growth but should be fine

### DIFF
--- a/src/util/process.rs
+++ b/src/util/process.rs
@@ -63,7 +63,7 @@ impl Communicator {
             Ok(())
         });
 
-        let buf = vec![0u8; 1 << 20];
+        let buf = vec![0u8; 1 << 10];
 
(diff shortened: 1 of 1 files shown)


keks (approved):
Looks good! I would also be happy with just growing if there is no space left anymore (`self.pos == self.buf.len()`), but this is fine, too.

PR #356, opened 2026-07-13
include oracle/claim in smt solver error messages (#198)

### LINKED ISSUE #198: Tell user which invariant file / randomness mapping file causes the error
When an SMT error (parsing error, tuple with wrong number of arguments etc.) occurs, then we could tell the user:
- Which SMT file the line number we give refers to.
- Which **user-generated** file might likely be the cause of the error, namely the last user file (invariant file or randomness mapping) which was loaded by the prover.

siccegge:


src/gamehops/equivalence/verify_fn.rs
```diff
@@ -194,23 +194,41 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
-                    self.backend.new_smtsolver()?
+                    self.backend.new_smtsolver()
                 }
-            };
+            }
+            .map_err(|err| Error::ProverProcessError {
+                claim_name: claim.name().to_string(),
+                oracle_name: oracle.name().to_string(),
+                solver_errors: vec![err],
+            })?;
```
keks: Can we have a

```rust
impl Error {
  fn prover_process_error(claim_name: &str, oracle_name: &str, err: crate::util::smtsolver::Error) -> Self {
    Self {
      claim_name: claim_name.to_string(),
      oracle_name: oracle_name.to_string(),
      solver_errors: vec![err],
    }
  }
}
```

and then we can call that method with simpler arguments here, looks nicer.

keks (commented):
Looks pretty good, just one small idea for making the code simpler.

keks (approved):
Thanks!

PR #357, opened 2026-07-14
Improve error reporting for matching equivalence exports

tornado80:


src/gamehops/equivalence/mod.rs
```diff
@@ -99,50 +99,119 @@ impl Equivalence {
...
+            unknown_oracle_names,
+        });
+    }
+
+    Ok(())
+}
+
 impl<'a> EquivalenceContext<'a> {
-    fn verify_exports_match(&self) -> (Vec<OracleSig>, Vec<OracleSig>) {
+    pub(crate) fn verify_exports_match(&self) -> Result<()> {
```
siccegge: Is there a reason to make this method public here?

src/gamehops/equivalence/mod.rs
```diff
@@ -99,50 +99,119 @@ impl Equivalence {
...
-            .cloned()
-            .collect();
-
-        (only_left, only_right)
+        verify_exports_match(
+            equivalence,
+            &left_game_inst.game().exports,
+            &right_game_inst.game().exports,
+        )
     }
```
siccegge, line 215: This method becomes trivial -- is there a reason we want to take `verify_exports_match` out as a standalone function?

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -47,26 +45,7 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
-                        .join(", "),
-                    export_difference
-                        .1
-                        .into_iter()
-                        .map(|sig| format!("{sig}"))
-                        .join(", "),
-                ),
-            });
-        }
+        self.eqctx.verify_exports_match()?;
```
siccegge, line 49: This is really the right thing -- making the verify function self-contained! THank you!

src/gamehops/equivalence/error.rs
```diff
@@ -35,11 +36,32 @@ pub enum Error {
...
     CompositionExportsMismatch {
         left_game_inst_name: String,
         right_game_inst_name: String,
-        mismatching_export_name: String,
+        /// oracle names (after aliasing) exported only by the left game instance
+        only_left: Vec<String>,
+        /// oracle names (after aliasing) exported only by the right game instance
+        only_right: Vec<String>,
+    },
+    #[error("oracle(s) {} are exported with different signatures by game instances {left_game_inst_name} and {right_game_inst_name}", .mismatches.iter().map(|mismatch| mismatch.oracle_name.as_str()).join(", "))]
```
siccegge, line 49: I learned a thing or two here about how to use Vec<>s in error messages :-D

src/gamehops/equivalence/mod.rs
tornado80: No, actually it should be private. Thanks.

src/gamehops/equivalence/mod.rs
tornado80: For easier unit testing of the possible error cases, I think it's good to pull out the core logic. Initially, I made it public for testing but then had to create an EquivalenceContext object. This also explains the left over public function.

src/gamehops/equivalence/verify_fn.rs
tornado80: I recently learned the "?" syntax as well and got excited to apply it right away :-D

siccegge (approved):

PR #358, opened 2026-07-14
Check induction start

### LINKED ISSUE #69: Check Induction Start
Right now we don't check the induction start for the invariant. Our language doesn't even have means to express what the base state should be.

- create a "starting state" which initializes all of the above values to their default and then check that the invariant holds
- optionally: allow initialization, have defaults for Int, Bool, Maybe, Table (?)
- check that "on the starting state", that the invariant holds and the randomness mapping is satisfiable
- Maybe add optional syntax: initialization block

### LINKED ISSUE #361: CLI: Allow the user to prove specific claims
Currently, we only allow the user to prove the claims of a specific oracle in a specific game hop of a theorem. However, There could be many claims in an oracle which makes it interesting to be prove individually. One technicality is that we have `<initial-state>` claims that are proved on an equivalence level rather than an oracle level. In future, we might have more lemmata to be proved on the equivalence level (or higher level) such as game invariants and package invariants or even custom user provided lemmata.

### LINKED ISSUE #370: Language: Empty bitstring literal
We are declaring empty bitstring constants for `Bits(*)` type. Let's expose a literal in the grammar for the user to use in the code of their packages.

Idea by @siccegge in #358

tornado80:
This PR checks state-relation, game, and package invariants hold in the induction start. Additionally,
- Introduces `empty-bitstring` literal for type `Bits(*)`
- Introduces `--invariant-start` option to CLI to be able to prove only induction-start claims. One can use `--claim invariant` to narrow down to the state-relation invariant or even `--claim package-invariant!Game-Package!` and `--claim game-invariant!Game!` to prove induction start only for the provided package and game invariants.
- Fixes a bug where countspec of the Bits(n) literals 0_n and 1_n are not resolved according to theorem constants.
- Refactors and unifies countspec resolution logic across the code base
- Gives a proper error message that CLI options`--invariant-start` and `--oracle` can not be used together. (Using `--oracle`skips induction start checks)
- Gives a proper error message when games to be proved equivalent do not expose any oracles to the adversary or their oracle sets do not match (name or signature-wise)
- Fixes invariants of 4WHS project to pass induction-start

Closes #69 and #370 and #361


src/gamehops/equivalence/verify_fn.rs
```diff
@@ -114,35 +114,86 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
-            .install(|| -> Vec<Result<()>> {
-                oracle_sequence
-                    .par_iter()
-                    .map(|oracle| -> Vec<Result<()>> {
-                        self.verify_oracle(ui.clone(), &smt, oracle)
-                    })
-                    .flatten()
-                    .collect()
-            })
+        let mut claims: Vec<_> = vec![self.verify_invariants_in_initial_state(ui.clone(), &smt)];
```
siccegge: Maybe move verification of the initial state inside the parallel solver?

You can "just" `.chain` a `std::iter::once`

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -114,35 +114,86 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
+            &self.eqctx.theorem().name,
+            &proofstep_name,
+            invariant_in_initial_state_claim_name,
+            1,
+        );
+
+        log::info!("verify: invariants in initial state");
+        // This is very nasty and we need to have only one set of
+        // invariants for the entire equivalence
+        // TODO: give good error instead of unwrap: empty oracle sequence, do we check this before?
```
siccegge: I think we just accept games if the adversary has no oracles so far -- which is fine, the adversary can't win if it has no oracle access :-D

In any case this is just a workaround because we haven't come around to moving invariants up to the equivalence and out of the individual oracles right?

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -114,35 +114,86 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
 
+    fn verify_invariants_in_initial_state<UI: TheoremUI + Send>(
+        &self,
+        ui: Arc<Mutex<&mut UI>>,
+        equivalence_smt: &[SmtExpr],
+    ) -> Result<()> {
+        let mut smt = vec![];
+        let eq = self.eqctx.equivalence();
+        let proofstep_name = format!("{} == {}", eq.left_name(), eq.right_name());
+        let invariant_in_initial_state_claim_name: &str = "invariants-in-initial-state";
```
siccegge: I might prefer something that doesn't clash with things the user might end up defining like say `<initial-state>` -- what do you think?

src/writers/smt/contexts/equivalence/emit.rs
```diff
@@ -241,6 +307,8 @@ impl<'a> EquivalenceContext<'a> {
         let mut base_declarations: Vec<SmtExpr> = vec![("set-logic", "ALL").into()];
 
         let mut bits_sort_suffixes = HashSet::new();
+        bits_sort_suffixes.insert("*".to_string());
+        base_declarations.extend(hacks::BitsDeclaration("*".to_string()));
```
siccegge: Why does the look directly below not already insert this?

src/writers/smt/contexts/equivalence/emit.rs
```diff
@@ -241,6 +307,8 @@ impl<'a> EquivalenceContext<'a> {
         let mut base_declarations: Vec<SmtExpr> = vec![("set-logic", "ALL").into()];
 
         let mut bits_sort_suffixes = HashSet::new();
+        bits_sort_suffixes.insert("*".to_string());
```
siccegge: Why do we need Bits(*) even if it's not used in the code?

src/writers/smt/contexts/equivalence/emit.rs
```diff
@@ -1205,6 +1275,56 @@ impl<'a> EquivalenceContext<'a> {
     }
 }
 
+/// The SMT value used to initialize a field of the given type in a game's initial state.
+/// Name of the opaque constant used as the "zero" value for Bits(*) fields. Declared once, on
+/// demand, by `emit_initial_state_values`.
+const BITS_ANY_ZERO_CONST: &str = "<0_*>";
```
siccegge: any reason to have an explicit constant for that? we don't do that for any of the other values!

src/theorem.rs
```diff
@@ -203,6 +203,7 @@ pub enum ClaimType {
     Lemma,
     Relation,
     Invariant,
+    InvariantInInitialState,
```
siccegge: Maybe InitialState would do? Just nitpiking here and might just be personal preference so feel free to ignore!

siccegge:
I guess we might want to bikeshed about the *name* of empty bitstring. Personally I find it a bit confusing to make it look like an all-zeros-string of arbitrary length as a concrete value :upside_down_face:

src/gamehops/equivalence/verify_fn.rs
tornado80: Apparently, `iter::once` takes the item but there is `std::iter::once_with` which accept a closure that can lazily return the item (which is what we need for parallelism). However, par_iter does not accept chained iterators with once_with. We can use `rayon::join` but that gives us a two `Result<()>` which we then need to merge into one list and check for failed claims. I think a better approach would be to make `verify_oracle` and `verify_invariants_in_initial_state` to return `IntoParallelIterator`'s to chain or a bunch of closures returning `Result<()>` which we can make an iterator out of them?

src/gamehops/equivalence/verify_fn.rs
tornado80: Good point. Here, we use it as a fake oracle name in the UI and it would appear as an oracle not a claim so I guess it's fine, but when generating smt transcript you are right and it can be confusing (still it is `LeftGame-RightGame-{invariants-in-initial-state}.smt2` while other claims are `LeftGame-RightGame-{OracleName}-{invariants-in-initial-state}.smt2`). I think user can not define oracles with "-" in their names as they are not smt_identifier in the grammar but they can surely have lemmata with such names. Later on claim name is used in verify_claim for UI as well as ClaimTheoremFailedError. So an idea is to add a field to Claim for DisplayName that we use UI-wise and that could be quite illustrative like "invariants-in-initial-state" while the actual name for SMT files could be `<initial-state>`? I also lean towards reserving these special claims (including later on `injective-randomness`) and we inform user they can not write claims with these names!

src/theorem.rs
tornado80: I like it. Unfortunately, I am by default verbose :)

src/writers/smt/contexts/equivalence/emit.rs
tornado80: Fair.

src/writers/smt/contexts/equivalence/emit.rs
tornado80: Fair.

src/writers/smt/contexts/equivalence/emit.rs
tornado80: Fair :)

src/gamehops/equivalence/verify_fn.rs
tornado80: Nevermind. Fixed now!

src/gamehops/equivalence/verify_fn.rs
tornado80: Actually, it's a very good idea to mark automatically inserted claims enclosed by angle brackets. This also applies to `<injective-randomness>`. Additionally, it could be `<same-output>`, etc. (if we make them optional to write if the user has no lemmata to write for them.)

tornado80:
How about `Bits_*_witness` instead of `0_*`? (inspired by EasyCrypt witnesses for types)

src/gamehops/equivalence/verify_fn.rs
tornado80: However, making verify_* functions to return claims allows us to filter claims and even allow user to choose claims!

siccegge: well you'll still need to also keep the context-specific smt (and you pay for smt export) if you first collect all the claims and then filter. And it could be "fun" to avoid copying the extra smt and getting it past the borrow checker -- but could be investigated

siccegge:
I think I prefer something where you don't need a strong verification background to not go WTF -- in the end what on earth is a witness (especially if you're not into intuistionistic logic). `<empty-bitstring>` ?

siccegge:
Also, there are some more empty/non/.. values all around the code, let's try to keep it somewhat consistent with naming. Also consider exposing the value to the package code -- it might be useful there as well.

Also might make sense to split this out as a separate pull request

src/gamehops/equivalence/verify_fn.rs
siccegge: The user might want to add dependencies to the same-output claim and therefore mention it in the lemma tree. The <> trick is neat because it's disallowed to use them in the context of theorem files.

On a side note, <> might not be too smart as you need something that is also valid as a filename on windows (I fell into that trap as well for package/game invariants, you can copy the solution from there!)

tornado80: Right. I am falling into this trap again and again :D Thank you.

src/gamehops/equivalence/verify_fn.rs
tornado80: Let's pull out the claim proving to #361.

src/gamehops/equivalence/verify_fn.rs
tornado80: Is the trick `!`? Do you think allowing claims to have an additional Display and/or File name would be better? Mainly, I don't like the `<initial-state>` in the UI and we have the file system issue as well.

tornado80:
Agreed. Let's decide on the empty bitstring literal in #370.

src/gamehops/equivalence/verify_fn.rs
tornado80: Now, the claim name can be anything and independent from what is chosen as file names.

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -93,35 +96,82 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
                     .collect()
-            })
+            });
+
+        let failed_claims = claims
             .into_iter()
             .filter_map(Result::err)
-            .collect();
-        if !failed_oracles.is_empty() {
+            .collect::<Vec<_>>();
```
siccegge: type signature needed?

siccegge:
Maybe @chrisbrzuska can take a look at the smt code we produce for checking induction start?

src/writers/smt/contexts/equivalence/emit.rs
```diff
@@ -40,10 +42,82 @@ impl<'a> EquivalenceContext<'a> {
...
+    }
+
+    fn emit_equivalence_claim_assert(&self, claim: &Claim) -> SmtExpr {
+        match claim.ty {
+            ClaimType::InitialState => self.emit_invariant_in_initial_state_assert(),
+            _ => unreachable!(),
+        }
+    }
+
+    fn emit_oracle_claim_assert(&self, claim: &Claim, oracle_name: &str) -> SmtExpr {
```
siccegge: scope_name ?

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -93,35 +96,79 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
+            &self.eqctx.theorem().name,
+            &proofstep_name,
+            claim_scope.name(),
+            1,
+        );
+
+        log::info!("verify: invariants in initial state");
+        // TODO: this is temporary workaround until we make the invariants equivalence-wide.
+        // For future: It's fine to unwrap for now as we accept games that don't expose any oracles.
+        let oracle_name = self.oracle_sequence().first().unwrap().name();
```
siccegge, line 202: I think we want something nicer in terms of error message when the user did not expose any oracles to the adversary **somewhere**

src/gamehops/equivalence/mod.rs
```diff
@@ -99,6 +99,21 @@ impl Equivalence {
     }
 }
 
+pub enum ClaimScope {
+    InitialState,
+    Oracle(String), // oracle name is stored
+}
```
keks, line 105: I would suggest a few changes here:
- let's document what this does
- not sure "Scope" describes this thing well
- I think if the variants had a little more data, then the iter-loop in verify_fn could iter over these enums, match it, and call the correct function, depending on context. If that is the structure, then this enum could be called InvariantClaim or so.

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -93,35 +96,79 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
         ui.lock().unwrap().proofstep_set_oracles(
             &self.eqctx.theorem().name,
             &proofstep_name,
-            oracle_sequence.len().try_into().unwrap(),
+            (oracle_sequence.len() + 1) // 1 is for checking invariants in the initial state
+                .try_into()
+                .unwrap(),
```
keks, line 101:
```suggestion
            oracle_sequence.len().into() + 1, // 1 is for checking invariants in the initial state
```

Turning a usize into a u64 should work on all relevant machines

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -93,35 +96,79 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
             .num_threads(self.parallel + 1) // one process is reserved for the "main" method
             .build()
             .unwrap()
             .install(|| -> Vec<Result<()>> {
-                oracle_sequence
-                    .par_iter()
-                    .map(|oracle| -> Vec<Result<()>> {
+                rayon::iter::once(())
+                    .map(|_| vec![self.verify_invariants_in_initial_state(ui.clone(), &smt)])
+                    .chain(oracle_sequence.par_iter().map(|oracle| -> Vec<Result<()>> {
```
keks, line 111: This is the loop I meant. If the items we iterate over are the enum above, extended with all the data needed to call `verify_invariants_in_initial_state` or `verify_oracle` respectively, then we could just chain the items and then do
```
  .map(|claim| match claim {
    InvariantClaim::Oracle(...) => self.verify_oracle(...),
    InvariantClaim::InitialState(...) => self.verify_invariants_in_initial_state(...),
  })
```

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -93,35 +96,79 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
+            claim_scope.name(),
+            1,
+        );
+
+        log::info!("verify: invariants in initial state");
+        // TODO: this is temporary workaround until we make the invariants equivalence-wide.
+        // For future: It's fine to unwrap for now as we accept games that don't expose any oracles.
+        let oracle_name = self.oracle_sequence().first().unwrap().name();
+        smt.append(&mut self.eqctx.emit_invariant(oracle_name));
+        smt.append(&mut self.eqctx.emit_initial_state_values());
```
keks, line 159: I am not quite sure what the workaround is, here, and what is meant with equivalence-wide.

src/writers/smt/contexts/equivalence/emit.rs
```diff
@@ -1205,6 +1284,47 @@ impl<'a> EquivalenceContext<'a> {
     }
 }
 
+fn default_smt_value(ty: &Type) -> SmtExpr {
```
keks, line 1287: I wonder if we could return an `Expression` here? That would seem a little more general.

keks (changes requested):
Looking forward to finally doing this properly, thanks :)

I have a few questions and suggestions on the Rust code. So far I have not reviewed the example and test projects.

chrisbrzuska (commented):
I looked into the SMT code generated for Full4WHS for the !INITIAL-STATE! check, and it matches how I expected it to look: In Tables, all entries are assigned to None, integers are assigned to 0.

I added a test_Bits_n_state project, and the !INITIAL-STATE! check sets it to 0_n which I also think is a natural choice.

The generated SMT code is also nicely modular and readeable.

For me, this all looks good.

Perhaps, I would like to have a look at the SMT code of Yao once, since Yao has package invariants and overall uses more complicated logic. I'll have a look at it after amir/induction-start has been re-based, so the bug-fixes for Yao from the main branch are there.

src/gamehops/equivalence/verify_fn.rs
tornado80: I think we discussed this in Retreat. Only for documenting, we only prove invariants of the "first" oracle of the equivalence holds in the initial state. The root cause is that we can define different invariants for each oracle which is a soundness issue and #365 addresses that.

src/gamehops/equivalence/verify_fn.rs
tornado80: Fair point. Apparently, indicatif progress bars require u64 and it has propagated to the UI trait which makes a conversion here necessary! I guess indicatify may not work on all machines then! What do you think? Maybe, we add it as a separate issue to to fix our machine arch dependency.

siccegge: it should be fixed in the ui trait -- the trait shouldn't leak indicatif internals -- but that's a separate step.

you can try jan's suggestion but I'd suspect it will just not compile

siccegge:
just noticxe3d i hvce some ancient "pending" commentss -- where they visible to people other than me?

src/gamehops/equivalence/verify_fn.rs
tornado80: Unfortunately, Jan's suggestion does not compile a requires explicit type annotation. Let's fix it in #406.

src/gamehops/equivalence/verify_fn.rs
tornado80: I think this is now resolved!

src/writers/smt/contexts/equivalence/emit.rs
tornado80: Thank you. Very good point. This duplicates SMT translation. The challenge was indeed in the Bits(*) which did not have a literal in the grammar. Now, we have `empty-bitstring` literal for `Bits(*)` similar to its SMT name.

tornado80:
@siccegge I couldn't see your previous comments. Only today I noticed them.

src/gamehops/equivalence/verify_fn.rs
tornado80: Right. Thanks for pinpointing. Now we give a nice error.

src/writers/smt/contexts/equivalence/emit.rs
tornado80: I dropped claim scope idea!

src/gamehops/equivalence/mod.rs
tornado80: Thanks for putting claim scope into spotlight. I think that was not a good idea and I had to refactor claim verification and UI trait.

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -99,129 +107,221 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
                 right_game_inst_name: eq.right_name.clone(),
-                failed_oracles,
+                failed_claims,
             });
         }
         Ok(())
     }
 
-    fn verify_oracle<UI: TheoremUI + Send>(
+    fn generate_game_or_package_invariant_induction_start_asserts(&self) -> Vec<(String, SmtExpr)> {
```
siccegge: this feels like it belongs into `writers::smt::context::equivalence::emit` what do you think?

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -47,10 +50,15 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
             req_oracle,
             req_claim,
             parallel,
+            only_induction_start,
         }
     }
 
     pub(crate) fn verify<UI: TheoremUI + Send>(&mut self, ui: &mut UI) -> Result<()> {
+        if self.only_induction_start && self.req_oracle.is_some() {
```
siccegge: is this the right location? Should it go into `crates/domino/src/main.rs` together with other checks on user arguments?

crates/domino/src/cli.rs
```diff
@@ -41,6 +41,8 @@ pub(crate) struct Prove {
     #[clap(short, long)]
     pub(crate) transcript: bool,
     #[clap(long)]
+    pub(crate) only_induction_start: bool,
```
siccegge: maybe `--induction-start` is enough? given that it's also `--claim` and not `--only-claim`

chrisbrzuska (commented):
My review is WIP: 
- I really like the way in which induction-start shows up in UI now (without !). 
- In the UI (and also as a claim), shall we rename induction-start to invariant-at-start? If someone is just trying to fix a proof which does not go through this might give a user more details regarding what they need to think about in case the user does not remember the full induction reasoning.
- I ran Full4WHS as "manual CI for large projects" and noticed that the last game-hop does not go through. I can have a look at the proof to see whether I have ideas how to make it go through (see attached file for the claims which failed).

Image

tornado80:
Thank you for the review. 
- Good suggestion. Shall we even call it `invariant-at-initial-state`?
- I also noticed the issue late evening yesterday and I mysteriously get stack overflow on the last hop. Does anyone else also get that?

chrisbrzuska:
> invariant-at-initial-state

That's more clear, but also slightly longer to type --- on the other hand, with the new *syntax for claims in the UI, length is not  really a problem anymore since --claim inv* would check invariant and invariant-at-initial-state which is nice.

Conclusion: I like your suggestion!

> stack overflow

As far as I can tell, I didn't get a stack overflow (OS Windows).

src/gamehops/equivalence/verify_fn.rs
tornado80: Yeah. I think main.rs makes more sense. Thanks.

crates/domino/src/cli.rs
tornado80: Yes. That's a good point. Thank you.

src/gamehops/equivalence/verify_fn.rs
tornado80: Good point. Let's keep SMTExpr generation based on claim type matching in emit module.

src/gamehops/equivalence/error.rs
```diff
@@ -97,10 +106,10 @@ pub enum Error {
     RewriteNeedsPackageContext { defn: String },
     #[error(transparent)]
     ParserError(#[from] crate::util::smtparser::Error),
-    #[error("SMT Solver failed in oracle {oracle_name}, claim {claim_name}")]
+    #[error("SMT Solver failed in claim {claim_name} of claim group {claim_group_name}")]
```
siccegge: is claim group something we want to expose to the user? it's fine as an internal name

src/gamehops/equivalence/verify_fn.rs
siccegge: hm but in fact it just moved further down now?

tornado80: Previously, part of the SMT generation logic was here but I moved all that logic to emit.rs but in verify_fn we drive SMT generation and deal with claims. Additionally, we have `generate_game_or_package_invariant_claims` which generates the claims for package and game invariants and I am reusing it so the claim names are consistent. When claims are extracted, we pass them to `emit_game_or_package_invariant_induction_start_assert` to get SMT code for each assertion but then assertions are checked in parallel later on in `verify_induction_start`. Otherwise, the logic of finding game and package invariants should move to emit.rs (which I don't think is the right place) or we generate package and game invariant claim names in two places independently which can risk divergence. We already regenerate the game and package invariant names independently in `emit_oracle_claim_assert` which might diverge. I think the distinction between the invariant itself and the claim would be solved with #392. What do you think?

src/gamehops/equivalence/error.rs
tornado80: I agree. What do you suggest as an alternative to claim group? I think oracle is misleading when the group is `invariant-at-initial-state`. Or should we be smart and know what type of claim group it is? We say nothing for `invariant-at-initial-state` but `oracle` for oracles?

src/gamehops/equivalence/error.rs
```diff
@@ -97,10 +106,10 @@ pub enum Error {
     RewriteNeedsPackageContext { defn: String },
     #[error(transparent)]
     ParserError(#[from] crate::util::smtparser::Error),
-    #[error("SMT Solver failed in oracle {oracle_name}, claim {claim_name}")]
+    #[error("SMT Solver failed in claim {claim_name} when verifying {claim_group_name}")]
```
siccegge, line 109: I like that!

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -6,6 +6,8 @@ use wildcard::Wildcard;
 use std::io::Write as _;
 use std::sync::{Arc, Mutex};
 
+use crate::gamehops::equivalence::verify_fn::ClaimGroup::{InductionStart, Oracle};
```
keks: What does this do? It looks like this is importing the variants of the enum defined in this current module?

crates/domino/src/main.rs
```diff
@@ -29,7 +29,17 @@ pub(crate) struct Cli {
...
+#[error("--oracle and --induction-start cannot be used together")]
+#[diagnostic(help(
+    "--induction-start restricts verification to the induction start, which \
+        doesn't involve any oracle, so --oracle has no effect there. \
+        Pass only one of the two options."
+))]
+pub struct ReqOracleWithOnlyInductionStart;
+
 #[allow(clippy::large_enum_variant)]
+#[allow(clippy::enum_variant_names)]
```
keks, line 42: Can we instead drop the "Error" at the end of the variant names?

src/writers/smt/expr_term.rs
```diff
@@ -86,7 +91,7 @@ impl From<Expression> for Term {
             ExpressionKind::Or(exprs) => theories::core::or(exprs),
             ExpressionKind::Xor(exprs) => theories::core::xor(exprs),
             ExpressionKind::Identifier(ident) => ident.into(),
-            ExpressionKind::Bot => "bot".into(),
+            ExpressionKind::Bot => "mk-empty".into(),
```
keks: I think Bot is pretty broken right now. Maybe just undo this and we figure out a correct fix later?

keks (commented):
I have not reviewed the claim structure very deeply, this is more a general code style and passing values around correctly kind of review. Just a few nits on that level.

src/writers/smt/expr_term.rs
tornado80: Let's panic then as we discussed :D

src/gamehops/equivalence/verify_fn.rs
tornado80: Fair.

crates/domino/src/main.rs
tornado80: Good point. Let's not silent clippy :D Thank you.

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -258,84 +367,107 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
+            return Ok(());
+        }
+
+        let mut smt = equivalence_smt.to_owned();
+        smt.append(&mut oracle_smt.to_owned());
+        smt.push(self.eqctx.emit_oracle_claim_assert(claim, oracle_name));
+        let claim_group = ClaimGroup::Oracle {
+            oracle_name: oracle_name.to_string(),
+        };
+        self.verify_with_solver(smt, &claim_group, claim.name(), oracle_name)
```
siccegge: this call is wrong

siccegge (changes requested):

src/gamehops/equivalence/verify_fn.rs
tornado80: My bad. Thanks a lot for spotting this!

chrisbrzuska:
Summary of my review:
- I added tests for game and package invariants.
- I now also checked the smtlib code for game and package invariants as well and it looks good.
- I worked on 4WHSFull to make induction start go through. This was successful, but I did not manage to make other parts of the proof go through.
- Amir and I discussed back and forth about naming and ended up agreeing on invariant-start instead of invariants-at-initial state and induction start in UI/CLI.

chrisbrzuska (approved):
Thanks a lot! Induction start is a very important soundness check, and I very much like the granular style you developed for it.

PR #359, opened 2026-07-15
clippy came to its senses

siccegge:


siccegge:
Ah nix is still 4 versions behind current stable (0.1.93 vs 0.1.97) so remains affected. But I guess we can still ignore that (nix fails currently anyhow and will catch up soon)

tornado80 (approved):

Issue #67, opened 2025-04-10
Figure out whether we can avoid universal quantifier for randomness relation.


chrisbrzuska:
Sometimes, CVC5 has trouble handling the universal quantifier in the assert saying that Rand-Relation(ctr,ctr',sam-id,sam-id',offset,offset') => Rand-Equality(ctr+offset,ctr'+offset',sam-id,-sam-id'). Can't we assume  Rand-Equality directly where we need it? This would avoid a universal quantifier. This suggestion is by @tornado80 .

keks:
```
cvc5 <<EOF                                                                                                              ✔
(declare-fun f (Int) Int)
(declare-fun g (Int) Int)
(assert (forall   ((x Int))        (= (f x) (g x))))
(assert (exists ((x Int)) (not (= (f x) (g x)))))
(check-sat)
EOF
unsat
```

PR #360, opened 2026-07-15
Quantifier free randomness mapping 

### LINKED ISSUE #206: Define randomness mappings without quantifiers?
We were pondering with @tornado80 whether it would be possible to define some/most randomness mappings without quantifiers, as instead of this (simplified version of) randomness mapping

```
(assert (not
  (forall (sample-id-left sample-id-right) 
    (=> 
      (and (= sample-id-left "foo") (= sample-id-right "bar"))
      (rand-is-eq sample-id-left sample-id-right)
     )
   )
))
```

This specific example can be simplified to:
```
(assert (not
   (rand-is-eq "foo" "bar")
))
```

The more common example of a randomness mapping would be in DNF-ish form (OR of ANDs: `(id-left == "foo" AND id-left == "bar") OR (id-left == "baz3" AND id-right == "bar")`); this should translate into multiple asserts of the above form.

Indeed there also the counters that one would need to somehow take into account, but I guess the most usual thing would be to restrict the allowed values of the new counter, based on the "old" one, for which there is a pre-existing variable `(<game-Gleft-rand-pkg_left-EVAL-sampleleft> <<game-state-gleft-old>>)`. So if the user would express `sample-ctr-right` as a function `f` of `sample-ctr-old-right`, the assertion becomes `(rand-is-eq "foo" "bar" (f sample-ctr-old-left) (f sample-ctr-old-right))`, if I read the code correctly. For example, to say that the predicate holds only for first sample of a well-known sample ID, the identity function `f(x) = x` should do.

With this framework, from a quick look, I think no quantifier should be needed at least for the "happy path" of defining randomness mappings. For those cases, also it would maybe be possible to define a JSON file or similar, where the user would enumerate the randomness mappings, possibly defining the counter mappings in SMT-LIB format, if the function is not identity (but instead e.g. `sample-ctr-old-right + 1` to map the second sample for one of the sides).

The benefit of not using quantifiers for the happy paths would be to get fewer (or potentially no) unknown responses from the solver, which would help users especially that are new to domino and randomness mappings. Eventually, when there are better outputs for SAT responses, it'd also be easier to show a user-friendly explanation that "randomness here could be different to randomness there, which leads to ability to distinguish the two".

Are there any well-known cases of randomness mappings which do not fall in to this "happy path" (which is: left sample ID equals some constant, right sample ID equals some constant, left sample ID equals some function of old left sample ID, and right sample ID equals some function of old right sample ID)?

tornado80:
Closes #206 and #67

### DIFF
--- a/src/transforms/samplify.rs
+++ b/src/transforms/samplify.rs
@@ -1,10 +1,10 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
 use crate::expressions::Expression;
-use crate::package::Composition;
+use crate::package::{Composition, Export};
 use crate::statement::{Assignment, AssignmentRhs, CodeBlock, IfThenElse, Pattern, Statement};
 use crate::types::Type;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::convert::Infallible;
 use std::iter::FromIterator;
 
@@ -12,7 +12,7 @@ use std::iter::FromIterator;
 
 pub struct Transformation<'a>(pub &'a Composition);
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct Position {
     pub game_name: String,
     pub inst_name: String,
@@ -32,6 +32,8 @@ pub struct SampleInfo {
     pub tys: Vec<Type>,
     pub count: usize,
     pub positions: Vec<Position>,
+    // maximum offset of each sampling operation for each exported oracle
+    pub max_offset: Option<HashMap<Export, HashMap<Position, usize>>>,
 }
 
 impl super::Transformation for Transformation<'_> {
@@ -45,7 +47,7 @@ impl super::Transformation for Transformation<'_> {
 
         let game_name = self.0.name.as_str();
 
-        let insts: Result<Vec<_>, Infallible> = self
+        let insts: Vec<_> = self
             .0
             .pkgs
             .iter()
@@ -70,16 +72,18 @@ impl super::Transformation for Transformation<'_> {
                 }
                 Ok(newinst)
             })
-            .collect();
+            .collect::<Result<Vec<_>, Infallible>>()?;
+
         Ok((
             Composition {
-                pkgs: insts?,
+                pkgs: insts,
                 ..self.0.clone()
             },
             SampleInfo {
                 tys: Vec::from_iter(samplings),
                 count: ctr,
                 positions,
+                max_offset: None,
             },
         ))
     }
--- a/src/transforms/theorem_transforms.rs
+++ b/src/transforms/theorem_transforms.rs
@@ -5,7 +5,7 @@ use std::{collections::HashSet, convert::Infallible};
 use crate::{theorem::GameInstance, types::Type};
 
 use super::{
-    deconstructinvoke, loopunroll,
+    deconstructinvoke, loopunroll, max_offset_extractor,
     resolveoracles::{self, ResolutionError},
     returnify, samplify, tableinitialize, treeify, type_extract, unwrapify, GameTransform,
     Transformation,
@@ -53,7 +53,7 @@ fn transform_game_inst(
     let (comp, _) = deconstructinvoke::Transformation(&comp)
         .transform()
         .expect("splitinvoke failed unexpectedly");
-    let (comp, samplinginfo) = samplify::Transformation(&comp)
+    let (comp, mut samplinginfo) = samplify::Transformation(&comp)
         .transform()
         .expect("samplify transformation failed unexpectedly");
     let (comp, _) = unwrapify::Transformation(&comp)
@@ -64,6 +64,7 @@ fn transform_game_inst(
         .unwrap_or_else(|ResolutionError(failed_oracle_stmts)| {
             panic!("error resolving oracles: {failed_oracle_stmts:?}")
         });
+    max_offset_extractor::Transformation(&comp, &mut samplinginfo).transform();
     let (comp, _) = returnify::TransformNg
         .transform_game(&comp)
         .expect("returnify transformation failed unexpectedly");
--- a/src/writers/smt/contexts/equivalence/emit.rs
+++ b/src/writers/smt/contexts/equivalence/emit.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet, HashSet};
+use std::{collections::{BTreeMap, BTreeSet, HashSet}, fmt::Display};
 
 use crate::{
     expressions::{Expression, ExpressionKind},
@@ -40,6 +40,15 @@ pub(crate) struct RandomnessMappingEntry {
     pub(crate) sample_id_right: SmtExpr,
     pub(crate) offset_left: usize,
     pub(crate) offset_right: usize,
+}
+
+impl Display for RandomnessMappingEntry {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, 
+            "left id: {}, right id: {}, left offset: {}, right offset: {}", 
+            self.sample_id_left, self.sample_id_right, 
+            self.offset_left, self.offset_right)
+    }
 }
 
 #[derive(Clone, Debug, Default)]
--- a/src/writers/smt/contexts/equivalence/emit.rs
+++ b/src/writers/smt/contexts/equivalence/emit.rs
@@ -13,7 +13,7 @@ use crate::{
     writers::smt::{
         contexts::{EquivalenceContext, GameInstanceContext, GenericOracleContext},
         declare::declare_const,
-        exprs::{SmtAnd, SmtAssert, SmtEq2, SmtExpr, SmtForall, SmtImplies, SmtIte, SmtNot},
+        exprs::{SmtAnd, SmtAssert, SmtEq2, SmtExpr, SmtImplies, SmtIte, SmtNot},
         patterns,
         patterns::{
             const_mapping::GameConstMappingFunction,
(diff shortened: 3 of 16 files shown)


src/transforms/samplify.rs
```diff
@@ -32,6 +32,8 @@ pub struct SampleInfo {
     pub tys: Vec<Type>,
     pub count: usize,
     pub positions: Vec<Position>,
+    // maximum offset of each sampling operation for each exported oracle
+    pub max_offset: Option<HashMap<Export, HashMap<Position, usize>>>,
```
siccegge, line 36: shouldn't this be `HashMap<Export, usize>` as we are already looking at a concrete sample position?

src/transforms/samplify.rs
```diff
@@ -70,16 +72,18 @@ impl super::Transformation for Transformation<'_> {
                 }
                 Ok(newinst)
             })
-            .collect();
+            .collect::<Result<Vec<_>, Infallible>>()?;
```
siccegge, line 75: ?

src/transforms/theorem_transforms.rs
```diff
@@ -64,6 +64,7 @@ fn transform_game_inst(
         .unwrap_or_else(|ResolutionError(failed_oracle_stmts)| {
             panic!("error resolving oracles: {failed_oracle_stmts:?}")
         });
+    max_offset_extractor::Transformation(&comp, &mut samplinginfo).transform();
```
siccegge, line 67: would it make more sense to fold this into samplify?

then you also can get rid of the Option

src/transforms/samplify.rs
tornado80: I think we have one SampleInfo for each game instance and not for each position. So the max_offset also computes the maximum possible counter for each sample position reachable from each exported oracle.

src/transforms/theorem_transforms.rs
tornado80: One issue is that max_offset_extractor needs to find all reachable sampling positions from an exported oracle so it needs oracle resolution. Samplify can be moved after oracle resolution. However, samplify and max_offset_extractor have different AST traversal order. Samplify linearly updates the code of oracles of all packages while max_offset_extractor DFS into each exported oracle. I think keeping them separate helps readability but we could move max_offset out of SampleInfo.

siccegge:
Ah one thing: it would be nice if we could keep the sample-infos in the order in which they appear in that export call. Then we can re-use the information also for a order-based randomness mapping!

I.e. instead of accumulating into a `HashMap`, collect into a `Vec` -- given that the number of entries is going to be small this might even be more efficient :upside_down_face:

tornado80:
Excellent point 👍

tornado80:
- Group by types!
- Assert equality of samplings instead of using rand-is-eq

src/transforms/theorem_transforms.rs
tornado80: Let's merge the transformations but keep the algorithms separate!

src/transforms/samplify.rs
```diff
@@ -156,6 +168,12 @@ pub fn samplify(
                 },
                 file_pos,
             ) => {
+                if !matches!(
+                    ty.kind(),
+                    TypeKind::Boolean | TypeKind::Integer | TypeKind::Bits(_)
+                ) {
+                    panic!("Only bits, bools, and integers are allowed for sampling");
+                }
```
keks, line 169: Do we catch that earlier and return an error or do we just panic if the user provides an invalid program? If this should be caught earlier, use `unreachable!` and add a test.

src/transforms/samplify.rs
```diff
@@ -199,6 +217,153 @@ pub fn samplify(
     Ok(CodeBlock(newcode))
 }
 
+type OffsetMap = Vec<(usize, usize)>;
```
keks: Just so I understand: The reason this is not a hashmap is so we maintain the ordering? Maybe the cleaner thing would be to have both a hashmap and a vec of keys that defines that order

src/transforms/samplify.rs
```diff
@@ -199,6 +217,153 @@ pub fn samplify(
...
+    }
+}
+
+// Concatenates the given maps and for elements that exist in both maps
+// compute the maximum offset.
+fn max_offsets(left: &OffsetMap, right: &OffsetMap) -> OffsetMap {
+    let mut result = left.to_owned();
+
+    for (pos, offset) in right {
+        if let Some(idx) = left.iter().position(|(p, _)| p == pos) {
```
keks: This is becoming O(n^2) now :D

src/transforms/samplify.rs
```diff
@@ -199,6 +217,153 @@ pub fn samplify(
...
+    let oracle = pkgs[pkg_idx]
+        .pkg
+        .oracles
+        .iter()
+        .find(|oracle| oracle.sig.name == oracle_name)
+        .unwrap_or_else(|| {
+            panic!(
+                "could not find oracle {oracle_name} in package instance {}",
+                pkgs[pkg_idx].name
+            )
```
keks: Should this be unreachable? then do unreachable! not panic!, and add a test that shows that this is caught earlier.

src/transforms/samplify.rs
```diff
@@ -199,6 +217,153 @@ pub fn samplify(
...
+            _,
+        ) => {
+            let edge = edge
+                .as_ref()
+                .unwrap_or_else(|| panic!("oracle invocation {oracle_name} is not resolved"));
+            oracle_offsets(pkgs, edge.to(), &edge.sig().name, positions)
+        }
+        Statement::InvokeOracle(invoke) => {
+            let edge = invoke.edge.as_ref().unwrap_or_else(|| {
+                panic!("oracle invocation {} is not resolved", invoke.oracle_name)
```
keks: same

src/transforms/samplify.rs
```diff
@@ -199,6 +217,153 @@ pub fn samplify(
...
+            oracle_offsets(pkgs, edge.to(), &edge.sig().name, positions)
+        }
+        Statement::IfThenElse(ite) if ite.else_block.0.is_empty() => {
+            codeblock_offsets(pkgs, pkg_idx, &ite.then_block, positions)
+        }
+        Statement::IfThenElse(ite) => max_offsets(
+            &codeblock_offsets(pkgs, pkg_idx, &ite.then_block, positions),
+            &codeblock_offsets(pkgs, pkg_idx, &ite.else_block, positions),
+        ),
+        Statement::For(..) => panic!("cannot extract sample max offset for loops"),
```
keks: again...

src/transforms/theorem_transforms.rs
```diff
@@ -45,17 +45,18 @@ fn transform_game_inst(
         .transform()
         .expect("type extraction transformation failed unexpectedly");
     /*
-     * Note: we currently do samplify early so a `if foo { stuff }
+     * Note 1: we currently do samplify early so a `if foo { stuff }
```
keks:
```suggestion
     * Note 1: we currently do samplify before treeify so a `if foo { stuff }
```

src/writers/smt/contexts/equivalence/emit.rs
```diff
@@ -985,18 +1118,14 @@ impl<'a> EquivalenceContext<'a> {
...
         {
             out.push(decl_ctr);
+            // the following could be merged together; it is important for
+            // randomness mapping to assert that old value is zero
             out.push(assert_ctr);
             out.push(assert_zero_ctr);
-            out.push(decl_val);
-            out.push(assert_val);
+            out.push(decl_val); // should not be needed
+            out.push(assert_val); // should not be needed
```
keks: remind me what these are. If they are not needed, why are they still here?

keks (commented):
I think this goes in the right direction, but I'd like to see some changes first.

src/transforms/samplify.rs
tornado80: Yes. We were thinking to use the per oracle ordering of samplings for auto randomness mapping but it already becomes a weird logic a bit down in max_offsets. max_offsets computes the maximum offsets of sample operations in the body of if and else branches. For max offset calculation, it is over approximation but for auto randomness mapping it might not make sense, i.e. automatically mapping the if cases and else cases! I think a hash map is nicer until we find the exact use case for order-based randomness mapping. We already have issue [#386](https://github.com/domino-lang/domino/issues/386) for name-based randomness mapping.

src/transforms/samplify.rs
tornado80: We don't catch them earlier. This is the first place and I noticed that we want to limit the types for sampling when I was trying to generate mapping candidates only for samplings with matching types. (We use `Types::types_match`) I think we might want to fix the Infallible error type for transforms to properly report these errors? Or is it conflicting with the new parser?

src/writers/smt/contexts/equivalence/emit.rs
tornado80: We extract randomness counters from the game states for each sampling operation. We also create constants for the first sampled value of each sample operations (i.e. ctr = 0). I am not sure we actually need any of these. The crucial part for randomness mapping precondition is that initial counters be zero so we don't need offsets. This does not necessarily need the initial counters to have a constant.

src/transforms/samplify.rs
tornado80: This could be much simpler with hash map. Shall I switch to hash map and figure out the ordering when we actually have the order-based auto rand mapping PR?

chrisbrzuska (approved):
I looked through the SMT code generated by (some oracles of):
- hello-world
- 4WHSFull
- YaoLayer
- kem-dem (ssp version & multi-challenge version)

The code looks like I would have expected. For Yao, the quadratic growth of the SMT code is suprisingly ok (= fast & still easy to read). In 4WHSFull, a game-hop fails, but that seems due to an unrelated issue.

Overall, the runs *felt* very fast :-). Thanks a lot!

Update: Running YaoLayer on another branch now. And the quantifier-free randomness mapping is *so* much faster :-).

src/transforms/sample_max_counter_extractor.rs
```diff
@@ -0,0 +1,585 @@
...
+fn add_offsets(target: &mut OffsetMap, source: &OffsetMap) {
+    for (pos, offset) in source {
+        *target.entry(*pos).or_insert(0) += offset;
+    }
+}
+
+// Takes union of the given maps and for elements that exist in both maps
+// compute the maximum offset.
+fn max_offsets(left: &OffsetMap, right: &OffsetMap) -> OffsetMap {
+    let mut result = left.to_owned();
```
keks:
```suggestion
    let mut result = left.clone();
```

let's call a spade a spade

src/writers/smt/contexts/equivalence/emit.rs
```diff
@@ -425,6 +424,123 @@ impl<'a> EquivalenceContext<'a> {
...
+            .unwrap_or_else(|| panic!("could not find right export {oracle_name}"));
+
+        let left_offsets = self
+            .max_offsets_left()
+            .get(left_export)
+            .unwrap_or_else(|| panic!("could not find max offsets for left export {oracle_name}"));
+        let right_offsets = self
+            .max_offsets_right()
+            .get(right_export)
+            .unwrap_or_else(|| panic!("could not find max offsets for right export {oracle_name}"));
```
keks: can we use unreachable! here?

src/writers/smt/contexts/equivalence/emit.rs
```diff
@@ -953,32 +1069,23 @@ impl<'a> EquivalenceContext<'a> {
...
         /////// randomess counters
 
-        for (decl_ctr, assert_ctr, assert_zero_ctr, decl_val, assert_val) in
-            build_rands(self.sample_info_left(), left)
-        {
+        for (decl_ctr, assert_ctr, assert_zero_ctr) in build_rands(self.sample_info_left(), left) {
             out.push(decl_ctr);
             out.push(assert_ctr);
+            // it is important for randomness mapping to assert that old counter is zero
+            // otherwise offset is needed
```
keks: I don't understand this, which may be because I forgot how exactly the smt mechanism works :sweat_smile: I think it would be nice if there was a comment block somewhere that explains this. We can do that in a followup, though

keks (approved):
Nice, looks good, just a few nits. Let's merge this tonight!

src/transforms/sample_max_counter_extractor.rs
tornado80: Right. Thank you. Forced me to read about the practice and the difference.

src/writers/smt/contexts/equivalence/emit.rs
tornado80: Yes, actually. Oracle resolution and sample_max_counter_extractor should have caught these. Thanks

src/writers/smt/contexts/equivalence/emit.rs
tornado80: Very good catch. It's because the comment is wrong😄

tornado80: So for offset-based randomness mapping, we assert that old counters are zero instead of adding the user-provided offset to the base counter!

PR #367, opened 2026-07-16
 default claims (#362) 

### LINKED ISSUE #362: Language: Make mandatory claims (same-output, equal-aborts, invariant) and oracles optional to write
In many gamehops of many proofs, we don't use lemmata to prove the core claims (same-output, equal-aborts, invariant) and it unnecessarily takes a lot of space of the theorem files. It is better to make writing them optional while they are automatically checked for the user. We could use angle brackets around the name of the claim (i.e. `<same-output>`) to illustrate that they are automatically generated.

Moreover, we should not require the user to write `no-abort` and it should be automatically inserted. However, the user can benefit from this helper in their own lemmata if need be.

siccegge:


tornado80 (approved):
Perfect. Very quick. Thank you. In the linked issue, I wrote that we shouldn't need to write no-abort but now I think if one can prove an oracle without no-abort, it's just stronger and if user is overriding it, it's good not to insert automatically and let the user prove a stronger claim?

siccegge:
We're only inserting default claims if they are not there yet -- so if the user wants to prove something different they are free to  say so and then do so ;-)

chrisbrzuska (approved):
That's nice. It will make proof files shorter and less noisy. It might also make it easier for users to start with Domino. Thanks @siccegge !

PR #372, opened 2026-07-16
[smtrewrite] also expose package consts

### LINKED ISSUE #387: 🪄 functions in package/game invariants
Make `pkg.function` work in smt code for package invariants in addition to the `<<func-function>>`

needs a different strategy but is easy to implement

siccegge:


src/gamehops/equivalence/smtrewrite.rs
```diff
@@ -53,6 +56,79 @@ impl<'a> SmtRewrite<'a> {
...
+
+fn generate_func_aliases_game<'a>(
+    game: &'a GameInstance,
+    prefix: String,
+) -> impl Iterator<Item = (SmtExpr, SmtExpr)> + use<'a> {
+    game.game
+        .pkgs
+        .iter()
+        .flat_map({
+            let prefix = prefix.clone();
```
keks, line 91: is this really needed? I don't think we need to clone for formatting

keks: If that doesn't work, maybe make prefix a &str, then it can be copied.

src/gamehops/equivalence/smtrewrite.rs
```diff
@@ -53,6 +56,79 @@ impl<'a> SmtRewrite<'a> {
...
+        .chain(game.consts.iter().filter_map(move |(ident, expr)| {
+            if matches!(ident.ty.kind(), TypeKind::Fn(_, _)) {
+                if let ExpressionKind::Identifier(thm_ident) = expr.kind() {
+                    let func_name = thm_ident.as_theorem_identifier().unwrap().ident_ref();
+                    Some((
+                        format!("{}.{}", prefix, ident.name).into(),
+                        format!("<<func-{func_name}>>").into(),
+                    ))
+                } else {
+                    unreachable!();
```
keks: Can we add a comment here that this is unreachable because we don't have function literals that could have been assigned instead of an identifier here?

Actually - What happens if we declare that there is a function `f: fn Int -> `fn Int -> Bool` and in the assignment for `g: fn Int -> Bool` we use `f(23)`? Pretty sure we don't support that, but very curious whether we return a proper error here or whether it fails in strange ways. If it would be passed on until here, it would panic.

src/gamehops/equivalence/smtrewrite.rs
```diff
@@ -53,6 +56,79 @@ impl<'a> SmtRewrite<'a> {
...
+    pkg.params.iter().filter_map(move |(ident, expr)| {
+        if matches!(ident.ty.kind(), TypeKind::Fn(_, _)) {
+            if let ExpressionKind::Identifier(thm_ident) = expr.kind() {
+                let func_name = thm_ident.as_theorem_identifier().unwrap().ident_ref();
+                Some((
+                    format!("{}.{}", prefix, ident.name).into(),
+                    format!("<<func-{func_name}>>").into(),
+                ))
+            } else {
+                unreachable!();
```
keks: same here

keks (commented):
Generally looks good. A few things where I am not sure if this may panic for strange code (when functions are used as values). Also some minor code cleanliness feedback.

src/gamehops/equivalence/smtrewrite.rs
siccegge: @keks I can't use &str because the borrow checker complains -- prefix can be a local variable inside `generate_func_aliases_game` and I can't borrow from that

src/gamehops/equivalence/smtrewrite.rs
siccegge:
```
    const f: fn Integer -> fn Integer -> Bool;
```

```
thread 'main' (1740260) panicked at src/writers/smt/exprs.rs:334:17:
not implemented: Fn([Type { kind: Integer }], Type { kind: Boolean })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

siccegge: let's make that an issue and fix it :upside_down_face:

siccegge: #390

keks:
So one part of this is that we declare package and game instances in our scope, which means that they must not collide with const names. I think I like that.

But maybe we should also make sure they don't collide with package names? Not sure.

PR #377, opened 2026-07-19
type inference for smtrewrite macros

siccegge:


siccegge:
I guess we need to loop at the kem/yao proofs and also adapt them (remove the types)

src/gamehops/equivalence/error.rs
```diff
@@ -80,7 +80,7 @@ pub enum Error {
         lemma_name: String,
         oracle_name: String,
     },
-    #[error("Expected 2-tuple (name type) for argument but got {argument}")]
+    #[error("Expected 1-tuple (name) for argument but got {argument}")]
```
tornado80: Can we make the error message more illustrative? It would be helpful to have invariant file name and/or game instance names.

src/gamehops/equivalence/error.rs
keks, line 1: nice :)

src/gamehops/equivalence/smtrewrite.rs
```diff
@@ -205,6 +229,11 @@ impl SmtParser<SmtExpr, Error> for SmtRewrite<'_> {
         args: Vec<SmtExpr>,
         body: SmtExpr,
     ) -> Result<SmtExpr> {
+        let equivalence_name = format!(
+            "{} = {}",
+            self.context.equivalence().left_name,
+            self.context.equivalence().right_name
+        );
```
keks: let's move this into some method, so the reader doesn't stumble over some variable that is only used in errors.

src/gamehops/equivalence/smtrewrite.rs
```diff
@@ -145,22 +158,25 @@ impl SmtParser<SmtExpr, Error> for SmtRewrite<'_> {
                     args.iter().map(|sexpr| format!("{sexpr}")).join(" ")
                 ),
                 expected: "2".to_string(),
+                equivalence: equivalence_name.clone(),
             });
         };
-        let [left_arg_name, _left_arg_type] = &left_arg[..] else {
+        let [left_arg_name] = &left_arg[..] else {
```
keks: let's drop the parens around left and right

keks (changes requested):
Really cool, just a few nits

tornado80:
Does it make sense to drop types for oracle arguments in lemmata?

PR #378, opened 2026-07-20
use theorem identifier for Bits_ for into::<SmtSort>()

siccegge:


src/packageinstance.rs
```diff
@@ -508,7 +508,7 @@ pub(crate) mod instantiate {
 
         pub(crate) fn rewrite_statement(&self, stmt: Statement) -> Statement {
             use crate::statement::{Assignment, AssignmentRhs, Pattern};
-            let type_rewrite_rules = self.base_rewrite_rules();
+            let _type_rewrite_rules = self.base_rewrite_rules();
```
keks: let's delete this (and the cascade of unused code that deleting this triggers)

keks (approved):
Thanks, good catch!

In the future, we maybe should rework this whole thing (and the expr map/mapfold functions) with the [Fold pattern].

[Fold pattern]: https://rust-unofficial.github.io/patterns/patterns/creational/fold.html

PR #380, opened 2026-07-21
Simplify 4WHS

tornado80:


example-projects/4WHS/theorem/simple/invariant-Hybrid0-Hybrid1.smt2
```diff
@@ -100,10 +63,4 @@
...
    (= left.KX.RevTested right.KX.RevTested)
 
-   (state= left right)
-
-   (keys-computed-correctly left right)
-
-   (time-of-nonces left right)
-   (time-of-sid left right)
-   (time-of-acceptance left right)))
+   (state= left right)))
```
siccegge, line 66: I think this needs a better name than state equality

siccegge:
I guess it would still be nice to have the simple proof stylistically somewhat similar -- but if the two middle functions from the second hop get hidden in package invariants it's essentially there (after finding a good name instead of state= ) it looks very nice

PR #385, opened 2026-07-22
Update Nix

keks:
This PR updates the flake.lock to newer versions and makes CI update the flake.lock (not persisted, just for testing during the run).


flake.nix
```diff
@@ -60,20 +60,19 @@
           cargoLock.lockFile = ./Cargo.lock;
           buildAndTestSubdir = "crates/domino/";
           cargoBuildFlags = [ "--workspace" ];
-          # TODO: The allow unused_assignments is there because of miette:
-          #       https://github.com/zkat/miette/issues/458
+          # NOTE: cargo check / clippy / fmt used to live here too. They are
+          # now separate checks (checks.cargo-check, checks.clippy,
+          # checks.treefmt) so they can be run independently and don't run
```
siccegge, line 65: i don't see the treefmt change is it hidden somewhere?

siccegge (approved):
I guess treefmt was already external and you just made clippy et all separate as well? no objections

flake.nix
keks: Maybe it was already there... It's on line 204

PR #392, opened 2026-07-31
split ParsedClaim / SmtClaim

### LINKED ISSUE #381: Soundness: warn if user defines arbitrary SMT code except for functions and macros
We should get warnings if the user make any assumption that is not verified by Domino similar to admitted claims.

### LINKED ISSUE #398: Dependency on same-output panics
In a code equivalence proof, we can assume lemmas which we proved before. In theory, we could also assume same-ouput, i.e., the theorem file could contain a line as follows:

                    invariant:    [no-abort, same-output:]

However, domino currently panics in this case, i.e., using same-output as a dependency does not work currently.

2nd part of the issue: If I understand correctly, domino looks for a claim and does not find it. In this case, domino could give a proper error message.

siccegge:


siccegge:
@chrisbrzuska if you have time you could give using this version a close look veryfy nothing looks fishy ;-)

it changes the logic on which claims to run quite a bit and the result **should** be identical :upside_down_face:

chrisbrzuska:
We can think about whether we prefer define-package-invariant Key over define-package-invariant-Key or vice versa. The current version is consistent with define-state-relation invariant, but my first attempt at writing named state relations used a dash instead of a space.

siccegge:
Note: I don't think smtclaim is the right name, open for suggestions

src/writers/smt/contexts/equivalence/emit.rs
```diff
@@ -173,31 +174,57 @@ impl<'a> EquivalenceContext<'a> {
...
+                    build_lemma_call.clone()(dep_name)
+                } else {
+                    let dep_claim = smt_claims
+                        .iter()
+                        .find(|claim| claim.name() == dep_name)
+                        .unwrap();
+                    match dep_claim.ty() {
+                        ClaimType::Lemma => build_lemma_call.clone()(dep_name),
+                        ClaimType::Relation => build_relation_call(dep_name),
+                        ClaimType::Invariant
```
tornado80, line 194: Apparently, it is useful to be able to assume all types of invariants on the new state. There is a test project here: [chris/test-for-panic](https://github.com/domino-lang/domino/tree/chris/test-for-panic)

src/gamehops/equivalence/smtclaim.rs
```diff
@@ -0,0 +1,123 @@
...
+#[diagnostic(code(domino::theorem::custom_smt), severity(Warning))]
+pub struct CustomSmtWarning {
+    pub smt: SmtExpr,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum ClaimType {
+    RawSmt,
+    Function,
+    //Randomness,
```
tornado80, line 17: Is the randomness a place holder for injectivity checks?

src/gamehops/equivalence/smtclaim.rs
```diff
@@ -0,0 +1,123 @@
...
+    Lemma,
```
tornado80, line 18: Shall we rename it to GeneralRelation as we discussed in retreat?

src/gamehops/equivalence/smtclaim.rs
```diff
@@ -0,0 +1,123 @@
...
+    Relation,
```
tornado80, line 19: And this to StateRelation?

src/writers/smt/contexts/equivalence/emit.rs
tornado80, line 154: Seems to be the same as `build_relation_call`?

src/gamehops/equivalence/smtclaim.rs
```diff
@@ -0,0 +1,123 @@
...
+                // some define-fun but not some useable claim type
+                return (ClaimType::Function, funname.to_string());
+            }
+        }
+        // not a define-fun, big warning!
+        eprintln!(
+            "{:?}",
+            miette::Report::new(CustomSmtWarning { smt: smt.clone() })
+        );
+        (ClaimType::RawSmt, String::default())
```
tornado80, line 58: I am not sure if we should categorize RawSmt and Function as ClaimType if they are not claims. Maybe `guess_from_smt` returns Option?

src/gamehops/equivalence/smtclaim.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct SmtClaim {
+    ty: ClaimType,
```
tornado80, line 64: Maybe SmtClaim could be SmtDef and then its type would be SmtDefType as follows:
```rust
pub enum SmtDefType {
    Raw,
    Function,
    Claim(ClaimType),
}
```
and we remove Raw and Function from ClaimType and we won't have unreachable when emitting code!

tornado80: I think RandomnessMapping macro can later on be part of this as well.

tornado80: `SmtItem` could be another name

src/gamehops/equivalence/smtclaim.rs
tornado80: This function could be move to `SmtDefType` which can return SmtDefType for all cases instead of Option. Suggested below!

src/gamehops/equivalence/smtclaim.rs
tornado80: I see, this is for randomness mapping function?

tornado80 (changes requested):
Thank you for this change. No more guess_from_name :D

src/gamehops/equivalence/smtclaim.rs
siccegge: yes this is for randomness mappings but I didn't fully implement it

src/gamehops/equivalence/smtclaim.rs
siccegge: yah I'm kind of not excited about having several levels of nested structs. It used to be an option for a while and it might get back there.

tornado80: I see. Then maybe it's better to have a flat repetition of ClaimTypes. I think it's a bit confusing to have say RandomnessMapping, SmtRaw and Function as claims.

PR #393, opened 2026-08-02
Remove superfluous admitted lemma for randomness for last hop in 4WHSfull.

chrisbrzuska:
I removed admitted randomness lemma from 4WHSFull, since it is not needed for cvc5. If you want to verify that it still works, this is the command:

domino prove --proof Full4WHS --proofstep 11


siccegge (approved):
Thank you!

PR #397, opened 2026-08-07
CLI: add --path to allow choosing a project path 

tornado80:
Add optional `--path` on sub commands `prove`, `proofsteps`, and `latex` to choose a Domino project by the path of the directory containing `ssp.toml`.

### DIFF
--- a/crates/domino/src/cli.rs
+++ b/crates/domino/src/cli.rs
@@ -14,7 +14,7 @@ pub(crate) enum Commands {
     /// Reformat file or directory
     Format(Format),
 
-    Proofsteps,
+    Proofsteps(Proofsteps),
 }
 
 #[derive(clap::Args, Debug)]
@@ -22,6 +22,10 @@ pub(crate) enum Commands {
 pub(crate) struct Format {
     /// Input to reformat
     pub(crate) input: Option<std::path::PathBuf>,
+    /// Path to the Domino project (or a subdirectory of it) to reformat when no input is given.
+    /// Defaults to searching the current directory and its ancestors for an `ssp.toml`.
+    #[clap(long, short = 'p')]
+    pub(crate) project: Option<std::path::PathBuf>,
 }
 
 #[derive(clap::Args, Debug)]
@@ -31,11 +35,19 @@ pub(crate) struct Latex {
     #[clap(short, long, default_value = "z3")]
     pub(crate) smtsolver: Option<SolverVariant>,
     // TODO: given we have a default here, it seems impossible to choose none
+    /// Path to the Domino project (or a subdirectory of it). Defaults to searching the current
+    /// directory and its ancestors for an `ssp.toml`.
+    #[clap(long, short = 'p')]
+    pub(crate) project: Option<std::path::PathBuf>,
 }
 
 #[derive(clap::Args, Debug)]
 #[clap(author, version, about, long_about = None)]
 pub(crate) struct Prove {
+    /// Path to the Domino project (or a subdirectory of it). Defaults to searching the current
+    /// directory and its ancestors for an `ssp.toml`.
+    #[clap(long, short = 'p')]
+    pub(crate) project: Option<std::path::PathBuf>,
     #[clap(short, long, default_value = "cvc5")]
     pub(crate) smtsolver: SolverVariant,
     #[clap(short, long)]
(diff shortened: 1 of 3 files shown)


siccegge:
seems to violate LLM policy ;-)

crates/domino/src/cli.rs
```diff
@@ -14,14 +14,18 @@ pub(crate) enum Commands {
...
 
 #[derive(clap::Args, Debug)]
 #[clap(author, version, about, long_about = None)]
 pub(crate) struct Format {
     /// Input to reformat
     pub(crate) input: Option<std::path::PathBuf>,
+    /// Path to the Domino project (or a subdirectory of it) to reformat when no input is given.
+    /// Defaults to searching the current directory and its ancestors for an `ssp.toml`.
+    #[clap(long, short = 'p')]
+    pub(crate) project: Option<std::path::PathBuf>,
```
siccegge: maybe path instead of project? else it sounds a bit like the oracle/claim/theorem filtering? also maybe don't use the `-p` short as that was prover until very recently

crates/domino/src/cli.rs
```diff
@@ -14,14 +14,18 @@ pub(crate) enum Commands {
...
-    Proofsteps,
+    Proofsteps(Proofsteps),
 }
 
 #[derive(clap::Args, Debug)]
 #[clap(author, version, about, long_about = None)]
 pub(crate) struct Format {
     /// Input to reformat
     pub(crate) input: Option<std::path::PathBuf>,
+    /// Path to the Domino project (or a subdirectory of it) to reformat when no input is given.
```
siccegge: is searching back from the provided path actually the right solution? Maybe the user should point domino at the root directory (and if --path was provided we skip the search for `ssp.toml`?

crates/domino/src/main.rs
```diff
@@ -40,19 +40,19 @@ enum Error {
     IncompatibleArgumentsErrorError(#[from] IncompatibleArgumentsError),
 }
 
-fn proofsteps() -> Result<(), Error> {
-    let project_root = project::directory::find_project_root()?;
+fn proofsteps(p: &Proofsteps) -> Result<(), Error> {
+    let project_root = project::directory::find_project_root(p.project.as_deref())?;
```
siccegge: if one goes with "the user needs to provide the correct path if they provide a path then this would probably become

```rust
let project_root = p.project.unwrap_or(project::directory::find_project_root());
```

or something along these lines

src/project/directory.rs
```diff
@@ -118,8 +116,15 @@ impl Project for DirectoryProject<'_> {
...
 }
 
-pub fn find_project_root() -> std::result::Result<std::path::PathBuf, super::error::Error> {
-    let mut dir = std::env::current_dir().map_err(FindProjectRootError::CurrentDir)?;
+/// Finds the root of the Domino project (the directory containing `ssp.toml`) by searching
+/// `start_dir` and its ancestors. If `start_dir` is `None`, searches from the current directory.
+pub fn find_project_root(
+    start_dir: Option<&Path>,
+) -> std::result::Result<std::path::PathBuf, super::error::Error> {
+    let mut dir = match start_dir {
```
siccegge: this should probably also be an unwrap_or()

src/project/directory.rs
```diff
@@ -118,8 +116,15 @@ impl Project for DirectoryProject<'_> {
     }
 }
 
-pub fn find_project_root() -> std::result::Result<std::path::PathBuf, super::error::Error> {
-    let mut dir = std::env::current_dir().map_err(FindProjectRootError::CurrentDir)?;
+/// Finds the root of the Domino project (the directory containing `ssp.toml`) by searching
+/// `start_dir` and its ancestors. If `start_dir` is `None`, searches from the current directory.
+pub fn find_project_root(
+    start_dir: Option<&Path>,
```
siccegge: is there a reason we pass a &Path instead of a PathBuf ?

as we probably don't need to have the path for other things we can transfer ownership to `find_project_root` and we skip converting between Path and PathBug

siccegge (changes requested):

crates/domino/src/cli.rs
tornado80: I see. Agreed. At the same time, if it at some point we allow compiling files, we may want to pass the path to specific Domino file? Do you think this makes sense?

tornado80: Let's think about it when we actually consider that! For now, explicit path skips the search for ssp.toml.

siccegge (approved):
Looks good now!

PR #402, opened 2026-08-14
cli rename proofstep -> gamehop, proof -> theorem

siccegge:


siccegge:
there is a point to be made that this could be cleaned up deeper into domino code :upside_down_face:

siccegge:
I promise to do the more thorough rename of internals when I adapt for multiple oracles/claims

tornado80 (changes requested):
Thanks for the name update :D 
I think modules`verify_fn`, `src::project::mod`, `src::ui::mod`, and `src::ui::indicatif` could benefit from name updates.

siccegge:
> I promise to do the more thorough rename of internals when I adapt for multiple oracles/claims

ja not going to do that before that change but we can also just wait until someone gets around to that

siccegge:
also important note: do not do that until all the other PRs touching this code are done

PR #403, opened 2026-08-17
periodic LaTeX updates

### LINKED ISSUE #422: LaTeX export broke.
Both 4WHSFull and 4WHSsimple have resolution issues, i.e., there is Rust code in the generated LaTeX:

Image

I don't know what the cause of the issue is, but one todo item for myself is to write tests for latex export which we can integrate into CI.

NB: 30 commits ago, it was already broken. 100 commits ago, it was still working.

siccegge:


chrisbrzuska (approved):
I ran the latex export on 4WHSSimple, 4WHSFull as well as another project, and the latex compiles and the resulting PDFs look good.

keks (commented):
looks good!

PR #404, opened 2026-08-17
Use a directory tree for storing smt files, include theorem name

siccegge:


src/gamehops/equivalence/verify_fn.rs
```diff
@@ -263,6 +263,7 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
                     let transcript_file: std::fs::File = self
                         .project
                         .get_joined_smt_file(
```
tornado80: Shall we give a better name to the function like `get_claim_smt_file` now that we don't have a single joined file?

src/project/mod.rs
```diff
@@ -230,19 +230,21 @@ pub trait Project {
...
-        path.push("_build/code_eq/joined/");
+        path.push("_build/code_eq/");
+        path.push(theorem_name);
+        path.push(format!("{left_game_name}-{right_game_name}"));
         std::fs::create_dir_all(&path)?;
 
-        path.push(format!(
-            "{left_game_name}-{right_game_name}-{oracle_name}-{claim_name}.smt2"
-        ));
+        path.push(format!("{oracle_name}-{claim_name}.smt2"));
```
tornado80: Do you think it makes sense for the oracle to have its own directory?

keks: I second that question without strong opinions in either direction. I guess if an oracle frequently has >10 claims then it makes sense. But if it usually is just 2 or 3, then it also seems overkill. not sure.

siccegge: between 3 and 60 I would say

src/gamehops/equivalence/verify_fn.rs
siccegge: it could. was trying to keep this PR minimal though ;-)

siccegge: is claim_smt_file the right name?

src/project/mod.rs
siccegge: and i guess for the interesting cases at least ~10 is quite common

tornado80 (approved):

PR #405, opened 2026-08-20
move `invariant: []` up one level

### LINKED ISSUE #365: Soundness: Make user to define only one invariant for all oracles of the equivalence
Currently, we allow the user to give the paths to SMT-LIB files containing their invariants, randomness mappings, lemmata, assertions, etc. for each oracle. This means the user can provide different invariants for each oracle which is a soundness bug. Moreover, it makes it very weird in the code to take invariants from one of the oracles to prove `<initial-state>`. One should be able to declare invariants once in the equivalence level. We could check with our SMT parser that only one invariant function is defined.

siccegge:


src/gamehops/equivalence/mod.rs
```diff
@@ -240,97 +225,87 @@ impl<'a> EquivalenceContext<'a> {
...
-                        pkg,
-                        &file_contents,
-                    )?);
-                }
-            }
-
-            // Load Game Invariants
-            for file_name in &left_gctx.game().invariants {
+        let mut out = Vec::new();
+        //let mut linter = lint::Linter::new(self, oracle_name);
```
keks: stray comment, but also -- we have a linter??

Ah, I see, this is old... anyway let's remove old code comments, also below.

keks (approved):
Looks really good! Let's just remove the old commented out code.

PR #408, opened 2026-08-24
actualy disable incremental mode

siccegge:
from cvc5 documentation:

```
i | incremental [type bool , default true ] (also --no-* )

    enable incremental solving
```


chrisbrzuska (approved):
This PR only adds argument --no-incremental to smt calls.

PR #409, opened 2026-08-25
Check injectivity of randomness mapping relation

### LINKED ISSUE #72: Check that the provided randomness mapping is injective
In order for a randomness mapping to be meaningful, it needs to at least be injective, i.e. if we say that the first sampled value in the left game is the same as more than one sampled value in the right game, then there an error should be returned.

We should be able to ask the SMT solver to do that for us. However, there may be complex mappings for which it is not able to produce a meaningful answer.

Discussions:
- [#SSBee > injectivity of randomness mapping](https://chat.cryptozoo.eu/#narrow/channel/40-SSBee/topic/injectivity.20of.20randomness.20mapping)

tornado80:
Check injectivity of randomness mapping. Injectivity of the user-provided randomness mapping relation is verified in parallel to other oracle claims as two sub claims: if the relation is injective wrt. the left and right (id, ctr) pairs. The claims are called `!injective-randmap-2-to-1!` and `!injective-randmap-2-to-1!` and can be selected with `--claim` option of CLI. Additionally, one can restrict the verification to injectivity check via `--injective-randmap` CLI option. We verify injectivity of the randomness mapping for arbitrary old states rather than restricting them to old states satisfying the invariant.

### DIFF
--- a/src/gamehops/equivalence/verify_fn.rs
+++ b/src/gamehops/equivalence/verify_fn.rs
@@ -6,6 +6,7 @@ use wildcard::Wildcard;
 use std::io::Write as _;
 use std::sync::{Arc, Mutex};
 
+use crate::theorem::RandomnessMappingInjectivityCheck;
 use crate::writers::smt::contexts::GameInstanceContext;
 use crate::{
     gamehops::equivalence::error::{ClaimTheoremFailedError, Error, Result},
@@ -14,10 +15,7 @@ use crate::{
     theorem::{Claim, ClaimType},
     ui::TheoremUI,
     util::smtsolver::{SmtSolver, SmtSolverBackend, SmtSolverResponse},
-    writers::smt::{
-        contexts::{EquivalenceContext, RandomnessMappingInjectivityCheck},
-        exprs::SmtExpr,
-    },
+    writers::smt::{contexts::EquivalenceContext, exprs::SmtExpr},
 };
 
 pub(crate) struct EquivalenceSmtDriver<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync> {
--- a/src/gamehops/equivalence/verify_fn.rs
+++ b/src/gamehops/equivalence/verify_fn.rs
@@ -129,6 +129,8 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
         );
         smt.append(&mut self.eqctx.emit_constant_declarations());
 
+        smt.append(&mut self.eqctx.emit_invariant());
+
         let proofstep_name = format!("{} == {}", eq.left_name(), eq.right_name());
         let oracle_sequence = self.oracle_sequence();
 
@@ -198,7 +200,6 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
         log::info!("verify: invariants at initial state");
 
         let mut base_smt = equivalence_smt.to_owned();
-        base_smt.append(&mut self.eqctx.emit_invariant());
         base_smt.append(&mut self.eqctx.emit_initial_state_values());
 
         let mut checks: Vec<(String, SmtExpr)> = vec![(
@@ -352,7 +353,6 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
 
         log::info!("verify: oracle:{oracle:?}");
         smt.append(&mut self.eqctx.emit_auto_randomness(oracle.name()));
-        smt.append(&mut self.eqctx.emit_invariant());
 
         self.verify_as_ui_claim_group(ui.clone(), &claim_group, num_claims, || {
             self.do_verify_oracle(ui.clone(), &smt, oracle, &claims, &claim_group)
--- a/src/gamehops/equivalence/verify_fn.rs
+++ b/src/gamehops/equivalence/verify_fn.rs
@@ -213,9 +213,14 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
                 .par_iter()
                 .filter(|(claim_name, _)| self.is_claim_requested(claim_name))
                 .map(|(claim_name, assert)| {
-                    let mut smt = base_smt.clone();
-                    smt.push(assert.clone());
-                    self.verify_with_solver_as_ui_claim(ui.clone(), &smt, &claim_group, claim_name)
+                    let claim_smt = vec![assert.clone()];
+                    self.verify_with_solver_as_ui_claim(
+                        ui.clone(),
+                        &base_smt,
+                        &claim_smt,
+                        &claim_group,
+                        claim_name,
+                    )
                 })
                 .collect()
         })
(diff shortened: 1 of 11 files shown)


src/gamehops/equivalence/smtrewrite.rs
```diff
@@ -510,6 +513,207 @@ impl SmtParser<SmtExpr, Error> for SmtRewrite<'_> {
...
+        };
+
+        let left_name = atom_name(left_arg)?;
+        let right_name = atom_name(right_arg)?;
+        let consts_name = atom_name(consts_arg)?;
+
+        let left_state_name = format!("{left_name}.state");
+        let right_state_name = format!("{right_name}.state");
+
+        let statebindings = vec![
```
siccegge: i think the important question here is whether we want to take the information from the global namespace or whether it's cleaner to make it an argument to the function.

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -339,35 +364,96 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
         oracle_name: &str,
         claim: &Claim,
+    ) -> Result<()> {
+        self.verify_claim_with_ui(ui, oracle_name, claim.name(), || {
+            self.do_verify_claim(equivalence_smt, oracle_smt, oracle_name, claim)
+        })
+    }
+
+    /// Runs `verify` for the given claim, telling the ui when it starts and when it finishes.
+    fn verify_claim_with_ui<UI: TheoremUI>(
```
siccegge: do we really want to make this a separate functino? especially as with #286 this becomes three lines total

siccegge: but maybe I just don't yet see the logic in which the functions are sliced up?

keks: I think this is a very reasonable function. Or did you mean `verify_claim`?

keks (commented):
Generally looks good, also the claim. I agree we could have another look at the boundaries of the verify_* functions, but not too unhappy with it.

src/gamehops/equivalence/verify_fn.rs
siccegge: I think I understand the structure now after @tornado80 explained it and it makes sense. (as referring to the stack of 3-line functions ;-)

src/gamehops/equivalence/smtrewrite.rs
tornado80: I have no strong opinions. This way we would be backward compatible. What do you think? @keks

src/theorem.rs
```diff
@@ -286,7 +303,7 @@ pub struct Theorem<'a> {
 }
 
 impl<'a> Theorem<'a> {
-    pub(crate) fn with_new_instances(&self, instances: Vec<GameInstance>) -> Theorem<'a> {
+    pub fn with_new_instances(&self, instances: Vec<GameInstance>) -> Theorem<'a> {
```
siccegge: hm why this change?

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -332,31 +337,39 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
         ui.lock().unwrap().start_claim_group(
             &self.eqctx.theorem().name,
             &proofstep_name,
             &claim_group.ui_name(),
-            claims.len().try_into().unwrap(),
+            num_claims.try_into().unwrap(),
         );
 
-        log::info!("verify: oracle:{oracle:?}");
-        smt.extend(&mut self.eqctx.emit_return_value_helpers(oracle.name()));
```
siccegge, line 340: where has the return value helper gone?

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -428,42 +409,39 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
         claim_group: &ClaimGroup,
     ) -> Result<()> {
-        self.verify_claim_with_ui(ui, claim_group, claim.name(), || {
-            self.do_verify_claim(equivalence_smt, oracle_smt, oracle_name, claim, claim_group)
-        })
+        if claim.is_admitted() {
+            return Ok(());
+        }
+
+        let mut smt = smt.to_owned();
```
siccegge: hm the previous version deliberately tried to avoid copying the (oracle-wide) smt for each claim while this reverts to creating a separate copy

src/gamehops/equivalence/verify_fn.rs
```diff
@@ -314,31 +299,23 @@ impl<'a, Backend: SmtSolverBackend + Sync, Proj: Project + Sync>
...
         RandomnessMappingInjectivityCheck::ALL
             .as_slice()
             .par_iter()
             .filter(|check| self.is_claim_requested(check.name()))
             .map(|check| {
                 let claim_name = check.name();
 
-                let mut smt = base_smt.clone();
+                let mut smt = smt.to_owned();
                 smt.extend(check.emit_randomness_mapping_injectivity_check(oracle_name));
```
siccegge: same here, do we really need to copy things?

src/theorem.rs
tornado80: Agreed. Not intended! Possibly during rewriting history when splitting flexible randomness mapping that touched transforms.Anyway, it would be a mistake even there.

src/gamehops/equivalence/verify_fn.rs
tornado80: Apparently, it moved to verify_oracle_claim

chrisbrzuska (approved):
- I reviewed the SMT code for both directions of the randomness injectivity and think that it is a very nice encoding of the injectivity.

It states that one cannot find two things on the left side which are:
- both in relation with a shared thing on the right side, and
- distinct.
(same for the other direction).

- CLI option is --injective-randmap which I think is intuitive.
- UI sais !injective-randmap-2-to-1! and !injective-randmap-1-to-2!, which is clear. The ! are not pretty, but this is something which we will address globally with other claims which share this.

Additionally, I ran several of the example-projects. I am also double-checking Yao now based on your comments @tornado80 , but that should not prevent the merge.

src/gamehops/equivalence/verify_fn.rs
tornado80: Now, all claims should share the base oracle smt. However, we still copy oracle smt once per oracle because we only need return value helpers for oracle claim verification but not the injectivity check. We copy equivalence smt once per claim group (oracle and invariant-start) because of the interface of the verify_with_solver_as_ui_claim function. To share a common solver verification function, we need to roughly be agnostic about the smt expressions. However, invariant-start needs equivalence_smt + claim_smt while oracle claims need equivalence_smt + oracle_smt (invariant, randomness mapping, helpers) + claim_smt.

src/gamehops/equivalence/verify_fn.rs
tornado80: Should be fixed now

keks (approved):
Looks good, let's get this in!

PR #412, opened 2026-08-26
simplify the name of lemmas to <{lemma-name}-{oracle-name}>

siccegge:


tornado80 (commented):
Perfect. Thanks. Can we also get rid of angle brackets? Also, what do you think about giving the oracle name its own position in the syntax? say `define-lemma case-i EVAL`?

Also, what do you think about hiding oracle arguemnts behind another arguments? i.e. `(old-left old-right return-left return-right 
args` then one can use `args.i`, ....
This way adding an argument does not need the user to update the lemmata!

A bit more extreme comment is that we already use old and return when naming the first four arguments for readability. what if we also hide them further? say `left.old.state`, `left.return.value`, `left.return.state`? Then it would be only three arguments for lemma: `(left right args)`

One could also put the arguments as `left.args` and `right.args` then we won't take the arguments of the left oracle implicitly but it might be overkill?

src/gamehops/equivalence/smtrewrite.rs
```diff
@@ -49,9 +50,12 @@ pub enum SmtStatementKind {
 impl SmtStatementKind {
     pub fn name(&self) -> String {
         match self {
-            SmtStatementKind::StateRelation { name }
-            | SmtStatementKind::Function { name }
-            | SmtStatementKind::GeneralRelation { name } => name.to_string(),
+            SmtStatementKind::StateRelation { name } | SmtStatementKind::Function { name } => {
+                name.to_string()
```
tornado80: Thank you for this change. Do we want to use similar naming for state relations as well?

siccegge: not sure what you mean?

tornado80: Good point that we want to write state relations and they should have a nice name.

PR #418, opened 2026-08-28
Allow flexible randomness mapping

### LINKED ISSUE #152: Randomness Mapping not Flexible Enough
For yao we want to express something along the lines of 

```quote
If the state bit is false then map left 3 to right 2, else map it to right 4
```

Equally plausible is depending on an input bit.

One can do this right now by accessing the globally bound <arg-??> and <gamestateOld> variables but (a) that feels unclean and (b) makes reasoning about whether the mapping is admissible quite hard!

tornado80:
Generalized randomness mapping relation using `define-randomness-mapping {ORACLE-NAME}` macro. This macro accepts three inputs say `(left right consts)` where the sampling id's and counters/offsets are accessible via `left/right.id` and `left/right.ctr`, the package states of the old game states are accessible through `left/right.state.PkgName.StateVarName`, the left and right oracle arguments are accessible via `left/right.args.ArgName` and finally theorem constants (both functions and other constants) are accessible through `consts.ConstName`. After #372 is merged, one can drop this argument and expose them through `left/right.ConstName` and/or `left/right.PkgName.ConstName`. See updated example projects. Note that old randomness mapping can coexist with the new macro as the macro creates bindings with respect to existing constants in the transcript instead of defining a new function that takes all these arguments.


siccegge (commented):
This is great! I think we need to come up with a way to make the `handle_define_*` functions a bit less unyielding and annoying, but that's not specific to this PR.

Let's finish discussion on how precisely the arguments are supposed to look and merge this

src/gamehops/equivalence/smtrewrite.rs
keks, line 1: This is very long, in part because we do for both the left and the right. Would there be a way to extract that part into a function and just call it twice?

keks (commented):
I think this looks good. My only feedback is that the visitor method is pretty repetitive, and I wonder if parts of it could be factored into a subfunction that is called for both left and right.

PR #419, opened 2026-08-28
Express KEM correctness as SMT assertion

tornado80:


chrisbrzuska (commented):
If exists also works, perhaps, this might be nicer than using ghost variables. I usually think of ghost variables as something which I want to use in a proof step, but would prefer not to have in the top-level definition. What do you think about this?

chrisbrzuska (approved):
Thanks! I think that this is a useful example project to showcase how to use custom smt code --- not that it's terribly recommended, but I think that it can be a shortcut which one can justify, and I think that it makes the proof structure easier in this case. When we have time, we can add an analogous proof which has a reduction to correctness.

PR #420, opened 2026-08-29
[smtrewrite] Parse Smt into top-level statements

### LINKED ISSUE #381: Soundness: warn if user defines arbitrary SMT code except for functions and macros

siccegge:


src/util/smtparser/smt.pest
```diff
@@ -21,7 +21,9 @@ ty = { "Int" | "Bool" | sexp }
 list = { "(" ~ sexp*~ ")" }
 
 // Big Or, Fallback Entries last!
-sexp = { sampleid | defun | define_game_invariant | define_package_invariant | define_state_relation | define_lemma | list | integer | boolean | stringlit | atom }
-sexps = { SOI ~ sexp+ ~ EOI }
+sexp = _{ sampleid | list | integer | boolean | stringlit | atom }
 
+stmt = { defun | define_game_invariant | define_package_invariant | define_state_relation | define_lemma | sexp }
```
keks, line 26: As discussed on the call, let's open a followup for not hard-coding the syntax of our macros in the grammar, so we can give better error messages.

src/gamehops/equivalence/smtrewrite.rs
```diff
@@ -13,11 +16,58 @@ use crate::writers::smt::patterns::datastructures::DatastructurePattern;
 use crate::gamehops::equivalence::error::{Error, Result};
 use itertools::Itertools;
 
+#[derive(Error, Diagnostic, Debug)]
+#[error("custom smt in invariant file:\n{expr}")]
+#[diagnostic(code(domino::theorem::custom_smt), severity(Warning))]
+pub struct CustomSmtWarning {
+    pub expr: SmtExpr,
+}
```
keks, line 24: I think it would be nice if we had spans here, but that can also be for the future

src/gamehops/equivalence/smtrewrite.rs
```diff
@@ -13,11 +16,58 @@ use crate::writers::smt::patterns::datastructures::DatastructurePattern;
...
+
+#[derive(Debug, Copy, Clone)]
+pub enum SmtStatementSort {
```
keks: I think I would call it "Kind" - (a) to not confuse it with SMT sorts and (b) because that is a common naming pattern for enums-that-describt-structs (e.g. std::io::ErrorKind), but not too important

src/gamehops/equivalence/smtrewrite.rs
```diff
@@ -140,21 +190,69 @@ impl SmtRewrite<'_> {
     }
 }
 
-impl SmtParser<SmtExpr, Error> for SmtRewrite<'_> {
-    fn handle_atom(&mut self, content: &str) -> Result<SmtExpr> {
-        Ok(SmtExpr::Atom(content.to_string()))
+impl SmtParser<SmtObject, Error> for SmtRewrite<'_> {
+    fn handle_atom(&mut self, content: &str) -> Result<SmtObject> {
+        Ok(SmtObject::Expression(SmtExpr::Atom(content.to_string())))
```
keks: We could give the Visitor trait associated types for Expr/Term and Statement, maybe more. Then we can different types (but with hard-coded categorization) for different handle_* functions. Then we don't need the SmtObject and the assertions which variants the returned values are.

src/util/smtparser/implementation.rs
```diff
@@ -85,106 +85,106 @@ where
...
 
             self.handle_sexp(parsed)?;
         }
 
         Ok(())
     }
 
-    fn rule_sexp(&mut self, p: &Pair<Rule>) -> Result<T, E> {
-        debug_assert_matches!(p.as_rule(), Rule::sexp);
-        let inner = p.clone().into_inner().next().unwrap();
```
keks, line 114: Huh, why are these gone?

keks (commented):
I think we could merge it this way, but I have some suggestions that might make it a little nicer.

src/gamehops/equivalence/smtrewrite.rs
siccegge: yes fully agree. I also want spans for "other stuff" in that part (and this is a step towards broadly introducing spans)

src/util/smtparser/implementation.rs
siccegge: the into_inner() stuff moved into the caller

chrisbrzuska (approved):
I checked that this branch indeed throws a warning with smt code if I try to sneak in illegal code, and it does. :-)

src/gamehops/equivalence/smtrewrite.rs
siccegge: figured out what you meant here. it's the right thing to do, on it

src/util/smtparser/model.rs
```diff
@@ -2,15 +2,19 @@ use super::{Result, SmtParser};
 use crate::util::smtmodel::{SmtModel, SmtModelEntry};
 use crate::writers::smt::exprs::SmtExpr;
 
-enum ModelExtractorState {
+#[derive(Debug)]
+pub(crate) enum ModelExtractorState {
     SmtExpr(SmtExpr),
     Integer(i32),
     Boolean(bool),
     // SmtModelEntry(SmtModelEntry),
```
keks: Should we remove this comment? 

```suggestion
```

keks (approved):
I think this look really good now, just one stray comment that I think we can remove.

PR #424, opened 2026-09-02
allow multiple game/package invariants

siccegge:


chrisbrzuska (commented):
I checked that the domino throws a warning containing custom SMT code when included in one of the invariant files.

siccegge (commented):
Still Broken, does not properly emit pkg invariant name

tornado80 (approved):
Looks good. Thank you.

PR #428, opened 2026-09-09
do not sort invariant file names

siccegge:


tornado80 (approved):
Looks good. Thank you.

keks (approved):
Huh, looks good

chrisbrzuska (approved):
I included SMT files in different orders, and depending on the order, they were included into the SMT in the same order as I listed them.

